PR #7, opened 2021-09-27
Test with Yugabyte

patriknw:


ddl-scripts/create_tables_yugabyte.sql
```diff
@@ -0,0 +1,26 @@
...
+  event_payload BYTEA NOT NULL,
+
+  meta_ser_id INTEGER,
+  meta_ser_manifest VARCHAR(255),
+  meta_payload BYTEA,
+
+  PRIMARY KEY(slice, entity_type_hint, persistence_id, sequence_number)
+);
+
+CREATE INDEX event_journal_slice_idx ON event_journal(slice, entity_type_hint, db_timestamp);
```
patriknw, line 26: compared to postgress the table definition is the same but index creation couldn't have "IF NOT EXISTS"

docker/docker-compose-yugabyte.yml
```diff
@@ -0,0 +1,46 @@
...
+      - yb-tserver-data-1:/mnt/tserver
+    command: [ "/home/yugabyte/bin/yb-tserver",
+               "--fs_data_dirs=/mnt/tserver",
+               "--start_pgsql_proxy",
+               "--rpc_bind_addresses=yb-tserver-n1:9100",
+               "--tserver_master_addrs=yb-master-n1:7100",
+               "--ysql_sequence_cache_minval=1"]
+    ports:
+      - "9042:9042"
+      - "5433:5433"
```
patriknw, line 38: running Yugabyte on a different port in case one would like to have both running locally

PR #40, opened 2021-10-19
Review questions and feedback

johanandren:
Not really for merging

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
@@ -58,6 +58,7 @@ import org.slf4j.Logger
     }
 
     def asFutureDone(): Future[Done] = {
+      // FIXME really worth specific impl vs asFuture().map(_ => Done)(ExecutionContexts.parasitic)
       val promise = Promise[Done]()
       publisher.subscribe(new Subscriber[Any] {
         @volatile private var subscription: Subscription = null
@@ -115,6 +116,7 @@ import org.slf4j.Logger
     statement.execute().asFuture().flatMap { result =>
       val resultPublisher: Publisher[A] =
         result.map((row, _) => mapRow(row))
+      // note: it's already indexed seq (vector) so toindexedSeq is essentially no op, not copy
       Source.fromPublisher(resultPublisher).runWith(Sink.seq[A]).map(_.toIndexedSeq)(ExecutionContext.parasitic)
     }
   }
@@ -261,8 +263,10 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger)(impli
           if (log.isDebugEnabled())
             log.debug("{} - DB call failed: {}", logPrefix, exc.toString)
           // ok to rollback async like this, or should it be before completing the returned Future?
+          // Comment: gut feeling is that a single transformWith { case ok => commit; case nok => rollback } would look nicer anyway
           rollbackAndClose(connection)
         }
+
 
         result.flatMap { r =>
           commitAndClose(connection).map { _ =>
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcSerializer.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcSerializer.scala
@@ -40,6 +40,7 @@ import akka.serialization.SerializerWithStringManifest
   }
 
   private def offsetToBinary(offset: TimestampOffset): Array[Byte] = {
+    // Note: this is the same offset storage mechanism we use in spanner, encoding timestamp + each seen pid+seqNr when needed (same timestamp)
     val str = new java.lang.StringBuilder
     str.append(offset.timestamp)
     // readTimestamp not used in serialized
--- a/core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
@@ -87,6 +87,8 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
 
   private val journalTable = journalSettings.journalTableWithSchema
 
+  // Note: wasn't the thing that we can't use transaction_timestamp() because start of transaction rather than actual commit
+  // what about clock_timestamp() or maybe statement_timestamp()?
   private val insertEventSql = s"INSERT INTO $journalTable " +
     "(slice, entity_type_hint, persistence_id, sequence_number, db_timestamp, writer, write_timestamp, adapter_manifest, event_ser_id, event_ser_manifest, event_payload) " +
     "VALUES ($1, $2, $3, $4, transaction_timestamp(), $5, $6, $7, $8, $9, $10)"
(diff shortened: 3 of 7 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -58,6 +58,7 @@ import org.slf4j.Logger
     }
 
     def asFutureDone(): Future[Done] = {
+      // FIXME really worth specific impl vs asFuture().map(_ => Done)(ExecutionContexts.parasitic)
```
patriknw, line 61: the difference is the "completed without first value". I think a publisher without any value is often used in the r2dbc api.

btw, I see that there is a bug in
```
if (promise.isCompleted)
```
it should be `!promise.isCompleted`

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -261,9 +263,11 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger)(impli
           if (log.isDebugEnabled())
             log.debug("{} - DB call failed: {}", logPrefix, exc.toString)
           // ok to rollback async like this, or should it be before completing the returned Future?
+          // Comment: gut feeling is that a single transformWith { case ok => commit; case nok => rollback } would look nicer anyway
```
patriknw, line 266: 👍

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -115,6 +116,7 @@ import org.slf4j.Logger
     statement.execute().asFuture().flatMap { result =>
       val resultPublisher: Publisher[A] =
         result.map((row, _) => mapRow(row))
+      // note: it's already indexed seq (vector) so toindexedSeq is essentially no op, not copy
```
patriknw, line 119: it's a "cast" because I preferred to have IndexedSeq in the result type rather than `Seq` from the `Sink.seq`

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcSerializer.scala
```diff
@@ -40,6 +40,7 @@ import akka.serialization.SerializerWithStringManifest
   }
 
   private def offsetToBinary(offset: TimestampOffset): Array[Byte] = {
+    // Note: this is the same offset storage mechanism we use in spanner, encoding timestamp + each seen pid+seqNr when needed (same timestamp)
```
patriknw, line 43: yes, the serializer here isn't really used, unless reading (source provider) from slices and storing offset with another projection impl than r2dbc

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -87,6 +87,8 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
 
   private val journalTable = journalSettings.journalTableWithSchema
 
+  // Note: wasn't the thing that we can't use transaction_timestamp() because start of transaction rather than actual commit
```
patriknw, line 90: There is no way to insert the timestamp after commit, it's always when the transaction or statement started. This is one of the main reasons  why we need the backtracking and deduplication.

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -134,11 +136,13 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
             connection.createStatement(insertEventSql)
           if (events.size == 1)
             bind(stmt, events.head)
-          else
+          else {
             // TODO this is not used yet, batch statements doesn't work stmt.bind().add().bind().execute()
+            // Question: in which way does it not work, exception or silent ignore of binds?
```
patriknw, line 141: don't fully remember but I think it was something very visible

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/QueryDao.scala
```diff
@@ -40,6 +40,8 @@ private[r2dbc] class QueryDao(settings: R2dbcSettings, connectionFactory: Connec
   private val currentDbTimestampSql =
     "SELECT transaction_timestamp() AS db_timestamp"
 
+  // Question: but all 4 different queries are turned to prepared statements anyway by the driver even though somewhat
+  // dynamically generated here?
```
patriknw, line 44: yes, it's just a String so the statement cache must be the full strings

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -232,6 +232,10 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
 
     def nextQuery(state: EventsBySlicesState): (EventsBySlicesState, Option[Source[EventEnvelope, NotUsed]]) = {
       val newIdleCount = if (state.rowCount == 0) state.idleCount + 1 else 0
+
+      // Note: Found this hard to decipher - start backtracking if we saw no new events in 5 polls and time between
+      // last seen timestamp and last backtracking timestamp is larger than half backtracking window (???)
+      // what is last backtracking timestamp?
```
patriknw, line 238: yup, and I'm not sure I have found the right decision criteria yet.

Currently, there are essentially two non-overlapping scans.
1. The normal that search from latest offset until now - 100ms.
2. The backtracking that search from latest backtracking offset until latest normal offset and now - 3sec.

Switching to backtracking if 5 idle (empty) results from previous queries, or if latest backtracking offset is too far bind latest normal offset (halfBacktrackingWindow).

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -302,6 +306,9 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
       nextQuery = nextQuery)
   }
 
+  // Note: very sneaky that this looks like a method but actually is a lambda that keeps state across invocations and
+  // also that it is called deserializeAndAddOffset but actually also filters out duplicates, can we make it something
+  // more obviously stateful while still testable in isolation (a Flow[SerializedJournalRow, EventEnvelope] perhaps)?
```
patriknw, line 311: Sounds good to improve that. I think it's a rather exact copy from spanner plugin. It's used in statefulMapConcat so it's not wrong that it has state.

projection/src/main/scala/akka/projection/eventsourced/scaladsl/EventSourcedProvider2.scala
```diff
@@ -20,7 +20,7 @@ import akka.projection.scaladsl.SourceProvider
 import akka.stream.scaladsl.Source
 
 // FIXME this should be incorporated in Akka Projections
-
+// Question: but cannot until eventsBySlice officially is part of Akka Persistence Query, right?
```
patriknw, line 23: right

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
patriknw: a quick improvement could be to have `def ..: Flow` and put the statefulMapConcat here together with the function

PR #41, opened 2021-10-19
code review 

octonato:
still walking through the code...

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/ContinuousQuery.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/ContinuousQuery.scala
@@ -49,6 +49,7 @@ private[r2dbc] object ContinuousQuery {
 final private[r2dbc] class ContinuousQuery[S, T](
     initialState: S,
     updateState: (S, T) => S,
+    // FIXME: can be simplified to S => Option[FiniteDuration] as S is never modified
     delayNextQuery: S => (S, Option[FiniteDuration]),
     nextQuery: S => (S, Option[Source[T, NotUsed]]))
     extends GraphStage[SourceShape[T]] {


core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -322,7 +322,7 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
         // has this already been seen?
         if (currentSequenceNrs.get(row.persistenceId).exists(_ >= row.sequenceNr)) {
           log.debugN(
-            "filtering [{}] [{}] as db timestamp is the same as last offset and is in seen [{}]",
```
octonato, line 325: Or maybe we want to say: it is in the 'seen' Map?

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -50,7 +50,7 @@ import org.slf4j.Logger
 
         override def onComplete(): Unit = {
           subscription = null
-          if (promise.isCompleted)
+          if (promise.isCompleted) // I guess what we need here is: if not (yet) completed, we complete with failure
```
octonato, line 53: ok, ignore. This has been already spotted in https://github.com/akka/akka-persistence-r2dbc/pull/40/files#r731926737

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -190,6 +191,10 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
       row
     }
 
+    // Question: select with no limit may be problematic if we materialize to Future
+    // can't we have a selectStream variant returning a Source?
+    // I'm assuming that the driver won't fetch everything in one shot,
+    // but publish following the pace of the subscriber
```
patriknw, line 197: There are two reasons for materializing eagerly to Future[Seq].

1) If we would return a Source it would be difficult to know when the connection can be closed. That must be done when all rows have been mapped (consumed).

2) Connections is a limited resource and it's better to return it to the pool as quickly as possible rather than having it open while the potentially slow consumer is processing all rows.

Limit is used.

core/src/main/scala/akka/persistence/r2dbc/journal/R2dbcJournal.scala
```diff
@@ -42,7 +42,10 @@ object R2dbcJournal {
  * INTERNAL API
  */
 @InternalApi
-final class R2dbcJournal(config: Config, cfgPath: String) extends AsyncWriteJournal {
+final class R2dbcJournal(
+    config: Config, // FIXME: config: Config not used
```
patriknw, line 46: I think it is needed for the Akka Persistence creation of the instance (via config). There are a few variants supported but not all combinations.

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -208,9 +210,9 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger)(impli
     }
   }
 
-  def selectOne[A](logPrefix: String)(statement: Connection => Statement, mapRow: Row => A): Future[Option[A]] = {
-    select(logPrefix)(statement, mapRow).map(_.headOption)
-  }
+  def selectOne[A](logPrefix: String)(statement: Connection => Statement, mapRow: Row => A): Future[Option[A]] =
+    // Suggestion: if mapRow returns `null`, we want to make it a None by default
```
patriknw, line 214: in that case we should use `null` as a filter also for mapRow in `select`, and then this wouldn't need to change?

octonato: Good point, the `select` could discard null values. That makes most sense when selecting a single value. Anyway, the `mapRow` defines what is `null`. Can be a that a whole `Row` is mapped to `null`. 

Another question is if we want to make it a general purpose public API for users. They could benefit of it when building projections, for example. Don't need to be a full blow framework around R2DBC, but some basic building blocks to make R2DBC play nice with Akka and Akka Streams. 

If we don't want that, those methods can be limited to the things we need internally.

PR #43, opened 2021-10-19
delayNextQuery is S => Opt[Duration]

octonato:
Simplification of delayNextQuery function


core/src/main/scala/akka/persistence/r2dbc/internal/ContinuousQuery.scala
```diff
@@ -49,7 +49,7 @@ private[r2dbc] object ContinuousQuery {
 final private[r2dbc] class ContinuousQuery[S, T](
     initialState: S,
     updateState: (S, T) => S,
-    delayNextQuery: S => (S, Option[FiniteDuration]),
+    delayNextQuery: S => Option[FiniteDuration],
```
patriknw, line 52: the intention was to wire the State in and out of all 3 functions, but we can add that back if we ever need to modify the state from delayNextQuery

patriknw (approved):
LGTM

Issue #46, opened 2021-10-20
transaction_timestamp() not as expected


patriknw:
Looks like the transaction_timestamp() is not monotonically increasing for the same persistenceId.

From a failing ChaosSpec run:

```
postgres=# select persistence_id, sequence_number, db_timestamp, write_timestamp from event_journal where persistence_id = 'TestEntity-1|p6' order by db_timestamp, sequence_number;

 persistence_id  | sequence_number |         db_timestamp          | write_timestamp
-----------------+-----------------+-------------------------------+-----------------
 TestEntity-1|p6 |               1 | 2021-10-20 14:55:50.54862+00  |   1634741750497
 TestEntity-1|p6 |               2 | 2021-10-20 14:55:50.699197+00 |   1634741750655
 TestEntity-1|p6 |               3 | 2021-10-20 14:55:50.750724+00 |   1634741750712
 TestEntity-1|p6 |               4 | 2021-10-20 14:55:56.860302+00 |   1634741756816
 TestEntity-1|p6 |               5 | 2021-10-20 14:56:01.528778+00 |   1634741761493
 TestEntity-1|p6 |               9 | 2021-10-20 14:56:03.613466+00 |   1634741763608
 TestEntity-1|p6 |               6 | 2021-10-20 14:56:03.637993+00 |   1634741763597
 TestEntity-1|p6 |               7 | 2021-10-20 14:56:03.645786+00 |   1634741763608
 TestEntity-1|p6 |               8 | 2021-10-20 14:56:03.647508+00 |   1634741763608
 TestEntity-1|p6 |              10 | 2021-10-20 14:56:10.347297+00 |   1634741770330
```

9 is before 6.

Here it was a PersistAll of 7,8,9 and that also shows that the transaction_timestamp() is not the same within the transaction, which is surprising.

We already have some FIXMEs for batch inserts and maybe this FIXME explains the "wrong" order within the transaction:
```
// FIXME is it ok to execute next like this before previous has completed?
```

However, more important is that 9 got an earlier timestamp than 6. That means that the `ORDER BY db_timestamp, sequence_number` in the slice query will not work.

This might also explain #36 

For reference, this is how it looks when ordered by the client timestamp:
```
postgres=# select persistence_id, sequence_number, db_timestamp, write_timestamp from event_journal where persistence_id = 'TestEntity-1|p6' order by write_timestamp, sequence_number;
 persistence_id  | sequence_number |         db_timestamp          | write_timestamp
[...]

Issue #33, opened 2021-10-18
Log warning: there is no transaction in progress


patriknw:
```
[WARN] [io.r2dbc.postgresql.client.ReactorNettyClient] [] [] [reactor-tcp-nio-1] - Notice: SEVERITY_LOCALIZED=WARNING, SEVERITY_NON_LOCALIZED=WARNING, CODE=25P01, MESSAGE=there is no transaction in progress, FILE=xact.c, 
```

For example when running ChaosSpec

patriknw:
I think this is related to the pool issue. Haven't seen it after disabling the pool.

PR #47, opened 2021-10-21
Transactions are wrong with the connection pool, #46

patriknw:
First I thought it was the auto-commit that was interferring, but that wasn't the case. Same without auto-commit.

* test that clearly reproduces the problem
* transaction_timestamp isn't the same within the transaction
  as it is supposed to be
* many warning logs about transactions, see issue #33
* Swithing back to the DummyConnectionPool from https://github.com/akka/akka-persistence-r2dbc/pull/22

References #46


patriknw:
I have run ChaosSpec and akka-projection-testing with this. Works much better now. We should of course find the root cause (issue #48) but things are at least working now.

Issue #29, opened 2021-10-18
ExceptionFactory$PostgresqlBadGrammarException: [26000] prepared statement "S_3" does not exist,4)


patriknw:
Shows up sporadically (not often)

patriknw:
https://github.com/akka/akka-persistence-r2dbc/runs/3948525838?check_suite_focus=true#step:7:157

PR #50, opened 2021-10-21
R2dbcExecutor RS-utilities (internal) public

johanandren:
This is all I need for the Akkasls view store as far as I can see.


patriknw:
test failure was the "PostgresqlBadGrammarException: prepared statement "S_2" does not exist", issue #29

Issue #31, opened 2021-10-18
Test Projection retry


patriknw:
Does the deduplication work for Projection retry (retry, not restart)?

PR #101, opened 2021-11-01
Make inflight more robust, #31

patriknw:
This is a follow up on https://github.com/akka/akka-persistence-r2dbc/pull/99

* The inflight Map(pid->seqNr) is needed for atLeastOnce or other projections
  that store the offset afterwards. When using the retry mechanism the Map should
  not be updated so that next retry can accept the envelope.
* Changed the isAccepted method to be pure (returning Boolean without side effect)
  and separate addInflight for the modification. That is better design in general,
  and makes it possible to only add to the inflight Map after successful processing
  instead of the revert mechanism that was implemented previously.
* Also combined isDuplicate and isAccepted into one method since they are always used
  together.

References #31

### DIFF
--- a/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -38,12 +38,6 @@ object R2dbcOffsetStore {
 
   final case class Record(pid: Pid, seqNr: SeqNr, timestamp: Instant)
 
-  sealed trait InflightEntry {
-    def seqNr: SeqNr
-  }
-  case class Processing(seqNr: SeqNr) extends InflightEntry
-  case class Recovering(seqNr: SeqNr) extends InflightEntry
-
   object State {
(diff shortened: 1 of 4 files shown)


projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -202,9 +197,10 @@ private[projection] class R2dbcOffsetStore(
   // we use AtomicReference and fail if the CAS fails.
   private val state = new AtomicReference(State.empty)
 
-  // transient state of inflight pid -> seqNr before they have been stored and included in `state`)
-  // this can be updated concurrently with CAS retries
-  private val inflight = new AtomicReference(Map.empty[Pid, InflightEntry])
+  // Transient state of inflight pid -> seqNr before they have been stored and included in `state`), which is
```
johanandren, line 200:
```suggestion
  // Transient state of inflight pid -> seqNr (before they have been stored and included in `state`), which is
```

johanandren (approved):

Issue #79, opened 2021-10-26
DurableStateStoreQuery


patriknw:
currentChanges and changes

Same offset tracking as for eventsBySlices.

DurableState should also be slice based.

Related to #78

PR #102, opened 2021-11-01
DurableState Projection, #79

patriknw:
References #79

### DIFF
--- a/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -18,6 +18,7 @@ import akka.Done
 import akka.actor.typed.ActorSystem
 import akka.actor.typed.scaladsl.LoggerOps
 import akka.annotation.InternalApi
+import akka.persistence.query.DurableStateChange
 import akka.persistence.query.UpdatedDurableState
 import akka.persistence.r2dbc.internal.R2dbcExecutor
 import akka.persistence.r2dbc.query.TimestampOffset
@@ -622,7 +623,10 @@ private[projection] class R2dbcOffsetStore(
             Record(change.persistenceId, change.revision, timestampOffset.timestamp),
             timestampOffset,
             strictSeqNr = false))
-      // FIXME case DeletedDurableState when that is added
+      case change: DurableStateChange[_] if change.offset.isInstanceOf[TimestampOffset] =>
+        // FIXME case DeletedDurableState when that is added
+        throw new IllegalArgumentException(
+          s"DurableStateChange [${change.getClass.getName}] not implemented yet. Please report bug at https://github.com/akka/akka-persistence-r2dbc/issues")
       case _ => None
     }
   }
--- a/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -18,6 +18,7 @@ import akka.Done
 import akka.actor.typed.ActorSystem
 import akka.actor.typed.scaladsl.LoggerOps
 import akka.annotation.InternalApi
+import akka.persistence.query.UpdatedDurableState
 import akka.persistence.r2dbc.internal.R2dbcExecutor
 import akka.persistence.r2dbc.query.TimestampOffset
 import akka.projection.MergeableOffset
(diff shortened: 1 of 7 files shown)


projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -452,81 +453,89 @@ private[projection] class R2dbcOffsetStore(
...
+
+      if (!ok)
+        logger.debug("Filtering out rejected sequence number (might be accepted later): {}", recordWithOffset)
+
+      ok
+    } else {
+      // strictSeqNr == false is for durable state where each revision might not be visible
+      val ok =
+        seqNr > currentInflight.getOrElse(pid, 0L) &&
+        seqNr > currentState.byPid.get(pid).map(_.seqNr).getOrElse(0L)
```
patriknw, line 508: For durable state changes we accept anything that has a greater revision than previously known revision.

projection/src/test/scala/akka/projection/r2dbc/R2dbcTimestampOffsetStoreSpec.scala
```diff
@@ -204,6 +218,9 @@ class R2dbcTimestampOffsetStoreSpec
       // and not less
       offsetStore.isAccepted(createEnvelope("p3", 4L, startTime, 10, "e3-4")) shouldBe false
       offsetStore.addInflight(env3)
+      // and then it's not accepted again
+      // FIXME the || logic in isAccepted is wrong
+      // offsetStore.isAccepted(env3) shouldBe false
```
patriknw, line 223: I'll fix this bug in a separate PR

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -597,4 +606,25 @@ private[projection] class R2dbcOffsetStore(
...
+            strictSeqNr = true))
+      case change: UpdatedDurableState[_] if change.offset.isInstanceOf[TimestampOffset] =>
+        val timestampOffset = change.offset.asInstanceOf[TimestampOffset]
+        Some(
+          RecordWithOffset(
+            Record(change.persistenceId, change.revision, timestampOffset.timestamp),
+            timestampOffset,
+            strictSeqNr = false))
+      // FIXME case DeletedDurableState when that is added
+      case _ => None
```
johanandren, line 626: Should that be a match error/exception rather and this method return `RecordWithOffset` directly? Because it is an error if we forget to implement it for DeletedDurableState when that is added, or if we get some non TimestampOffset envelope here.

projection/src/main/scala/akka/projection/state/scaladsl/DurableStateSourceProvider2.scala
```diff
@@ -0,0 +1,80 @@
...
+        durableStateStoreQuery
+          .changesBySlices(entityTypeHint, minSlice, maxSlice, offset)
+      }
+
+    override def extractOffset(stateChange: DurableStateChange[A]): Offset = stateChange.offset
+
+    override def extractCreationTime(stateChange: DurableStateChange[A]): Long =
+      stateChange match {
+        case u: UpdatedDurableState[_] => u.timestamp
+        case _                         => 0L // FIXME handle DeletedDurableState when that is added
```
johanandren, line 77: Same here, fail instead?

projection/src/test/scala/akka/projection/r2dbc/DurableStateEndToEndSpec.scala
```diff
@@ -0,0 +1,251 @@
...
+      handlers.foreach { case (projectionId, handler) =>
+        (0 until numberOfEntities).foreach { p =>
+          val persistenceId = PersistenceId(entityTypeHint, s"p$p")
+          val slice = DurableStateSourceProvider2.sliceForPersistenceId(
+            system,
+            R2dbcDurableStateStore.Identifier,
+            persistenceId.id)
+          withClue(s"projectionId $projectionId, persistenceId $persistenceId, slice $slice: ") {
+            if (handler.sliceRange.contains(slice)) {
+              eventually {
```
johanandren, line 232: ok, so eventually we see the highest revision, but the earlier updates may be "folded" into fewer than one per update.

johanandren (approved):
LGTM but wonder about the handling of unexpected envelopes

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
patriknw: yes, I'll change that

projection/src/test/scala/akka/projection/r2dbc/DurableStateEndToEndSpec.scala
patriknw: exactly

Issue #81, opened 2021-10-26
CurrentPersistenceIdsQuery


patriknw:
currentPersistenceIds (including the new paging api)

PR #105, opened 2021-11-02
persistenceIds query, #81

patriknw:
* for event sourced and durable state
* probably not very good performance of these queries but we can start like this

References #81

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
@@ -34,7 +34,7 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
     extends ReadJournal
     with CurrentEventsBySliceQuery
     with EventsBySliceQuery
-    with PersistenceIdsQuery {
+    with CurrentPersistenceIdsQuery {
   import R2dbcReadJournal.PersistenceIdsQueryState
 
   private val log = LoggerFactory.getLogger(getClass)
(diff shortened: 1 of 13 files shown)


core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/R2dbcDurableStateStore.scala
```diff
@@ -120,4 +126,45 @@ class R2dbcDurableStateStore[A](system: ExtendedActorSystem, config: Config, cfg
       offset: Offset): Source[DurableStateChange[A], NotUsed] =
     bySlice.liveBySlices("changesBySlices", entityTypeHint, minSlice, maxSlice, offset)
 
+  override def currentPersistenceIds(afterId: Option[String], limit: Long): Source[String, NotUsed] =
+    stateDao.persistenceIds(afterId, limit)
+
+  def currentPersistenceIds(): Source[String, NotUsed] = {
```
johanandren, line 132: Isn't the no-parameter one only for the event journal?

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -86,4 +91,45 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+      }
+    }
+
+    ContinuousQuery[PersistenceIdsQueryState, String](
+      initialState = PersistenceIdsQueryState(0, 0, ""),
+      updateState = updateState,
+      delayNextQuery = _ => None,
+      nextQuery = state => nextQuery(state))
+      .mapMaterializedValue(_ => NotUsed)
+  }
```
johanandren, line 134: So here ContinousQuery isn't really used for tailing but for fetching batches rather than _every_ id at once?

johanandren (approved):
LGTM with two questions.

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/R2dbcDurableStateStore.scala
patriknw: Yes, but I'm adding both for completeness. `currentPersistenceIds()` is for fetching all and `currentPersistenceIds(afterId)` is for paging.

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
patriknw: yes, that is right. Same approach as for `currentEventsBySlices` and `currentEventsByPersistenceId` will be the same

PR #107, opened 2021-11-02
Include slice in offset table, #57

### LINKED ISSUE #57: Include slice in the offset table
Easer change number of projections (change slice ranges) if we include the slice number in the offset table. Then a new projection with a different slice range can continue with the same offset rows without having to copy and such.

When deleting old records we should probably keep 1 from each slice.

One small issue is that the maxNumberOfSlices is a property on the write side, and for this we would have to recompute it from the persistenceId on the projection side. That would be difficult if we make maxNumberOfSlices configurable since the offset store isn't really aware of the source provider. Having a fixed maxNumberOfSlices might be ok? Alternatively the slice must be included in the EventEnvelope and then we need another EventEnvelope class.

patriknw:
Draft because on top of other PRs, but otherwise ready for review.

Fixes #57

### DIFF
--- a/ddl-scripts/create_tables_postgres.sql
+++ b/ddl-scripts/create_tables_postgres.sql
@@ -53,6 +53,7 @@ CREATE TABLE IF NOT EXISTS akka_projection_offset_store (
 CREATE TABLE IF NOT EXISTS akka_projection_timestamp_offset_store (
   projection_name VARCHAR(255) NOT NULL,
   projection_key VARCHAR(255) NOT NULL,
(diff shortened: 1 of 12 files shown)


johanandren (approved):
LGTM

I guess the keep-one-record-per-slice fixme can be a separate thing to investigate

patriknw:
> I guess the keep-one-record-per-slice fixme can be a separate thing to investigate

yes, created https://github.com/akka/akka-persistence-r2dbc/issues/115 for additional testing

PR #108, opened 2021-11-02
Create the test table from the tests

patriknw:
* previous problem was because the wrong connection management

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
@@ -128,6 +128,43 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger)(impli
     connectionFactory.create().asFuture()
 
   /**
+   * Run DDL statement with auto commit.
+   */
+  def executeDdl(logPrefix: String)(statement: Connection => Statement): Future[Done] = {
+    val startTime = System.nanoTime()
+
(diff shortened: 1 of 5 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -127,6 +127,43 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger)(impli
...
+  def executeDdl(logPrefix: String)(statement: Connection => Statement): Future[Done] = {
+    val startTime = System.nanoTime()
+
+    getConnection().flatMap { connection =>
+      connection.setAutoCommit(true).asFutureDone().flatMap { _ =>
+        val done =
+          try {
+            val stmt = statement(connection)
+            stmt.execute().asFuture().flatMap { result =>
+              result.getRowsUpdated.asFutureDone()
```
patriknw, line 142: somewhat surprising but it seems like ddl statements doesn't return the number of rows so when I tried to use the ordinary `updateOne` it failed with:
```
Publisher [MonoHandleFuseable] completed without first value.
```

johanandren: I think we got tripped elsewhere previously on the fact that Monos are more like `Future[Option[T]]` than `Future[T]`.

johanandren (approved):

PR #112, opened 2021-11-03
feat: Snapshot store added 

johanandren:
References #34

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
@@ -9,13 +9,13 @@ import akka.actor.typed.ActorSystem
 import akka.annotation.InternalApi
 import akka.dispatch.ExecutionContexts
 import akka.persistence.r2dbc.R2dbcSettings
-import akka.persistence.r2dbc.internal.{R2dbcExecutor, SliceUtils}
-import akka.persistence.{SelectedSnapshot, SnapshotMetadata, SnapshotSelectionCriteria}
-import akka.serialization.{Serialization, Serializers}
-import io.r2dbc.spi.{ConnectionFactory, Row}
-import org.slf4j.{Logger, LoggerFactory}
+import akka.persistence.r2dbc.internal.{ R2dbcExecutor, SliceUtils }
+import akka.persistence.{ SelectedSnapshot, SnapshotMetadata, SnapshotSelectionCriteria }
+import akka.serialization.{ Serialization, Serializers }
+import io.r2dbc.spi.{ ConnectionFactory, Row }
+import org.slf4j.{ Logger, LoggerFactory }
 
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.{ ExecutionContext, Future }
 import scala.util.Failure
 
 /**
--- a/core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
@@ -15,7 +15,6 @@ import io.r2dbc.spi.{ ConnectionFactory, Row }
 import org.slf4j.{ Logger, LoggerFactory }
 
 import scala.concurrent.{ ExecutionContext, Future }
-import scala.util.Failure
 
 /**
  * INTERNAL API
@@ -81,7 +80,7 @@ private[r2dbc] final class SnapshotDao(
   private val snapshotTable = settings.snapshotsTableWithSchema
   private val r2dbcExecutor = new R2dbcExecutor(connectionFactory, log)(ec, system)
 
-  private val insertSql =
+  private val upsertSql =
     s"""INSERT INTO $snapshotTable (
           slice,
           entity_type_hint,
@@ -104,7 +103,7 @@ private[r2dbc] final class SnapshotDao(
           meta_ser_manifest = null
         """
 
-  private val insertSqlMeta =
+  private val upsertWithMetaSql =
     s"""INSERT INTO $snapshotTable (
           slice,
           entity_type_hint,
@@ -195,8 +194,8 @@ private[r2dbc] final class SnapshotDao(
     val slice = SliceUtils.sliceForPersistenceId(metadata.persistenceId, settings.maxNumberOfSlices)
 
     val insert =
-      if (metadata.metadata.isEmpty) insertSql
-      else insertSqlMeta
+      if (metadata.metadata.isEmpty) upsertSql
+      else upsertWithMetaSql
 
     val snapshot = value.asInstanceOf[AnyRef]
     val serializedSnapshot = serialization.serialize(snapshot).get
--- a/core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
@@ -1,0 +1,278 @@
+/*
+ * Copyright (C) 2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.r2dbc.snapshot
+
+import akka.actor.ExtendedActorSystem
+import akka.actor.typed.ActorSystem
+import akka.annotation.InternalApi
+import akka.dispatch.ExecutionContexts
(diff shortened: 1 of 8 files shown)


core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
```diff
@@ -0,0 +1,278 @@
...
+ * Class for doing db interaction outside of an actor to avoid mistakes in future callbacks
+ */
+@InternalApi
+private[r2dbc] final class SnapshotDao(
+    settings: R2dbcSettings,
+    connectionFactory: ConnectionFactory,
+    serialization: Serialization)(implicit ec: ExecutionContext, system: ActorSystem[_]) {
+  import SnapshotDao._
+
+  private val snapshotTable = settings.snapshotsTable
```
patriknw, line 64: settings.snapshotsTableWithSchema

ddl-scripts/create_tables_postgres.sql
```diff
@@ -23,6 +23,22 @@ CREATE TABLE IF NOT EXISTS event_journal(
 
 CREATE INDEX IF NOT EXISTS event_journal_slice_idx ON event_journal(slice, entity_type_hint, db_timestamp);
 
+CREATE TABLE IF NOT EXISTS snapshot(
+  slice INT NOT NULL,
```
patriknw, line 27: yes, good to include the `slice` to have the same kind of primary key even though we don't have any plan to use it for snapshots

core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
```diff
@@ -0,0 +1,278 @@
...
+       } else "") +
+      (if (criteria.maxTimestamp != Long.MaxValue) {
+         paramIdx += 1
+         s" AND write_timestamp <= $$$paramIdx"
+       } else "") +
+      (if (criteria.minTimestamp != 0L) {
+         paramIdx += 1
+         s" AND write_timestamp >= $$$paramIdx"
+       } else "") +
+      " ORDER BY sequence_number DESC"
```
patriknw, line 137: if we are only interested in one row we should include `LIMIT 1`

core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
```diff
@@ -0,0 +1,278 @@
...
+          }
+          if (criteria.minTimestamp > 0L) {
+            bindIdx += 1
+            statement.bind(bindIdx, criteria.minTimestamp)
+          }
+          statement
+        },
+        row =>
+          Serialization.withTransportInformation(system.classicSystem.asInstanceOf[ExtendedActorSystem]) { () =>
+            deserializeSnapshotRow(row, serialization)
```
patriknw, line 169: To be able to release the connection as quickly as possible it would be better to deserialize outside of the db connection by first transforming the row to `SerializedSnapshotRow` with the byte array intact.

We have a similar issue in the event replay that I intend to adjust.

core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
```diff
@@ -0,0 +1,278 @@
...
+      val slice = SliceUtils.sliceForPersistenceId(metadata.persistenceId, settings.maxNumberOfSlices)
+
+      val insert =
+        if (metadata.metadata.isEmpty) insertSql
+        else insertSqlMeta
+
+      val snapshot = value.asInstanceOf[AnyRef]
+      val snapshotSerializer = serialization.findSerializerFor(snapshot)
+      val snapshotManifest = Serializers.manifestFor(snapshotSerializer, snapshot)
+      val serializedSnapshot = snapshotSerializer.toBinary(snapshot)
```
patriknw, line 187: in the journal we have
```
          val serialized = serialization.serialize(event).get
          val serializer = serialization.findSerializerFor(event)
          val manifest = Serializers.manifestFor(serializer, event)
          val id: Int = serializer.identifier
```

and then the `Serialization.withTransportInformation` isn't needed

core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
```diff
@@ -0,0 +1,278 @@
...
+                .bind(4, metadata.timestamp)
+                .bind(5, serializedSnapshot)
+                .bind(6, snapshotSerializer.identifier)
+                .bind(7, snapshotManifest)
+
+            serializedMeta.foreach { case (serializedMeta, serializerId, serializerManifest) =>
+              statement
+                .bind(8, serializedMeta)
+                .bind(9, serializerId)
+                .bind(10, serializerManifest)
```
patriknw, line 215: don't you have to bind those to null when there is no meta? see journal

core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
```diff
@@ -0,0 +1,278 @@
...
+            }
+
+            statement
+        }
+        .map(_ => ())(ExecutionContexts.parasitic)
+        .andThen { case Failure(ex) =>
+          ex.printStackTrace()
```
patriknw, line 222: leftover?

core/src/test/scala/akka/persistence/r2dbc/snapshot/R2dbcSnapshotStoreSpec.scala
```diff
@@ -0,0 +1,17 @@
...
+import akka.actor.typed.ActorSystem
+import akka.actor.typed.scaladsl.adapter._
+import akka.persistence.CapabilityFlag
+import akka.persistence.r2dbc.{ TestConfig, TestDbLifecycle }
+import akka.persistence.snapshot.SnapshotStoreSpec
+
+class R2dbcSnapshotStoreSpec extends SnapshotStoreSpec(TestConfig.config) with TestDbLifecycle {
+  def typedSystem: ActorSystem[_] = system.toTyped
+
+  protected override def supportsMetadata: CapabilityFlag = true
```
patriknw, line 16: do we need two tests, one with true and one with false?

core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
johanandren: Thanks, was wondering why I needed that when it wasn't needed elsewhere.

core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
johanandren: When I set null I got some error, so there's two different statements instead.

core/src/test/scala/akka/persistence/r2dbc/snapshot/R2dbcSnapshotStoreSpec.scala
johanandren: Enabling just adds one more test case to the existing so I think that this is fine?

ddl-scripts/create_tables_postgres.sql
johanandren: That was what I was thinking, for sharding to work as well in a distributed db (without having to have different DAO impl)

patriknw:
@johanandren could you look into https://github.com/akka/akka-persistence-r2dbc/issues/113 also?

johanandren:
#113 implemented as well

core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
```diff
@@ -0,0 +1,293 @@
...
+private[r2dbc] final class SnapshotDao(
+    settings: R2dbcSettings,
+    connectionFactory: ConnectionFactory,
+    serialization: Serialization)(implicit ec: ExecutionContext, system: ActorSystem[_]) {
+  import SnapshotDao._
+
+  private val snapshotTable = settings.snapshotsTableWithSchema
+  private val r2dbcExecutor = new R2dbcExecutor(connectionFactory, log)(ec, system)
+
+  private val insertSql =
```
patriknw, line 84: maybe rename to `upsert`

patriknw (approved):
LGTM, just a small rename suggestion

PR #120, opened 2021-11-05
Lookup timestamp of previous sequence number, #74

### LINKED ISSUE #74: How to handle first event with unknown sequence number
We must be careful with accepting events with seqNr > 1 where previous sequence number is not known by the offset store. It could have missed previous sequence number because how timestamp based offset works (see [design notes](https://github.com/akka/akka-persistence-r2dbc/blob/main/design-notes.md#offset)).

Current implementation will therefore reject such event unless it is 3 seconds old (configurable by `accept-new-sequence-number-after-age`). The idea is that it will be picked up by the backtracking query and then it would be safe since the transaction_timestamp issues would have been resolved.

I don't like that solution. It introduces additional delay for that case (an entity that wasn't active for a long time). How long must the delay be to be safe?

What can we do instead? I'm not sure, but one thought would be to query the journal for the db_timestamp of the previous event. If that is older than the offset store window then it's safe to accept immediately.

patriknw:
This became rather invasive but I'm quite convinced that it's needed.

* When the offset store receives an unknown sequence number
  in `isAccepted` it will try to lookup the timestamp of the
  previous event. If that is older than then time window of
  the offset store it will be accepted, otherwise rejected.
* The reason for rejecting is that it previous event could
  be missing when reading from the tail based on the latest
  timestamp offset.
* Backtracking will pick up the missing event and fill the
  gap.
* This is better than the previous accept-new-sequence-number-after-age
  (3 seconds) because it will not introduce that delay for the
  "normal case" when a new event is emitted and previous event
  was older than the offset store time window.
* Also, previous solution had the the uncertainty about if
  3 seconds would be enough and backtracking.behind-current-time
  can now be increased without adding to the delay mentioned above.

Fixes #74

### DIFF
--- a/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -20,12 +20,14 @@ import akka.actor.typed.scaladsl.LoggerOps
 import akka.annotation.InternalApi
 import akka.persistence.query.DurableStateChange
 import akka.persistence.query.UpdatedDurableState
+import akka.persistence.query.scaladsl.EventTimestampQuery
 import akka.persistence.r2dbc.internal.R2dbcExecutor
 import akka.persistence.r2dbc.internal.SliceUtils
 import akka.persistence.r2dbc.query.TimestampOffset
 import akka.projection.MergeableOffset
 import akka.projection.ProjectionId
 import akka.projection.eventsourced.EventEnvelope
+import akka.projection.eventsourced.scaladsl.TimestampOffsetBySlicesSourceProvider
 import akka.projection.internal.ManagementState
 import akka.projection.internal.OffsetSerialization
 import akka.projection.internal.OffsetSerialization.MultipleOffsets
@@ -120,6 +122,8 @@ object R2dbcOffsetStore {
   }
 
   val FutureDone: Future[Done] = Future.successful(Done)
+  val FutureTrue: Future[Boolean] = Future.successful(true)
+  val FutureFalse: Future[Boolean] = Future.successful(false)
 }
 
 /**
@@ -128,8 +132,7 @@ object R2dbcOffsetStore {
 @InternalApi
 private[projection] class R2dbcOffsetStore(
     projectionId: ProjectionId,
-    minSlice: Int,
-    maxSlice: Int,
+    sourceProvider: Option[TimestampOffsetBySlicesSourceProvider],
     system: ActorSystem[_],
     settings: R2dbcProjectionSettings,
     r2dbcExecutor: R2dbcExecutor,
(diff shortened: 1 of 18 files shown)


projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -507,43 +544,57 @@ private[projection] class R2dbcOffsetStore(
...
+            if (previousTimestamp.isBefore(currentState.latestTimestamp.minus(settings.timeWindow)))
+              true
+            else {
+              logger.info("Rejecting unknown sequence number (might be accepted later): {}", recordWithOffset)
+              false
+            }
+          case None =>
+            // previous not found, could have been deleted
+            true
+        }
```
johanandren, line 585: Not important but map could perhaps be on parasitic ec

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -184,17 +182,17 @@ private[projection] object R2dbcProjectionImpl {
...
-              Done
-            }
+        offsetStore.filterAccepted(envelopes).flatMap { acceptedEnvelopes =>
+          if (acceptedEnvelopes.isEmpty) {
+            FutureDone
+          } else {
+            delegate
+              .process(acceptedEnvelopes)
+              .map { _ =>
+                acceptedEnvelopes.foreach(offsetStore.addInflight)
```
johanandren, line 192: Not really related to this PR but worth adding a single `addInflight` for multiple envelopes instead of one CAS per envelope?

projection/src/test/scala/akka/projection/r2dbc/R2dbcTimestampOffsetProjectionSpec.scala
```diff
@@ -726,39 +723,29 @@ class R2dbcTimestampOffsetProjectionSpec
...
-        R2dbcProjection.groupedWithinAsync(
-          projectionId,
-          Some(settings),
-          sourceProvider = backtrackingSourceProvider(envelopes2),
-          handler = () => handler())
+        R2dbcProjection.groupedWithinAsync(projectionId, Some(settings), sourceProvider2, handler = () => handler())
 
       projectionTestKit.run(projection2) {
-        // FIXME result1 is "e1-1|e1-2|e1-3|e1-2|e1-3|e1-4|e1-5|e1-6|
+        // FIXME
```
johanandren, line 742: Is this a different problem or would it also be fixed by this PR?

johanandren (approved):
LGTM

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
patriknw: I think it's doing too much (risk for introducing something that could throw exception in the future), and this case shouldn't be used often so not a hot perf path

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
patriknw: ok, adding to my small fixes todo list

projection/src/test/scala/akka/projection/r2dbc/R2dbcTimestampOffsetProjectionSpec.scala
patriknw: I tried to enable it but it's still failing so will investigate separately (it's for grouped async case so not top prio)

PR #125, opened 2021-11-05
Remove linebreaks in sql

patriknw:
* because they may be logged (and it wasn't consistent in all places)


johanandren (approved):
LGTM 

Could've made sense to remove them programmatically just like stripMargin, or even only when logging but this is fine.

PR #130, opened 2021-11-05
Remove tags, since not implemented

patriknw:


core/src/main/scala/akka/persistence/r2dbc/journal/R2dbcJournal.scala
```diff
@@ -96,9 +96,9 @@ private[r2dbc] final class R2dbcJournal(config: Config, cfgPath: String) extends
     def atomicWrite(atomicWrite: AtomicWrite): Future[Try[Unit]] = {
       val serialized: Try[Seq[SerializedJournalRow]] = Try {
         atomicWrite.payload.map { pr =>
-          val (event, tags) = pr.payload match {
-            case Tagged(payload, tags) => (payload.asInstanceOf[AnyRef], tags)
-            case other                 => (other.asInstanceOf[AnyRef], Set.empty[String])
+          val event = pr.payload match {
+            case Tagged(payload, _) => payload.asInstanceOf[AnyRef] // tags not implemented, issue #82
```
johanandren, line 100: Should it perhaps make more noise or even fail rather than silently discard?

johanandren (approved):

core/src/main/scala/akka/persistence/r2dbc/journal/R2dbcJournal.scala
patriknw: ok, I'll add a "first usage" warning log

patriknw: added in https://github.com/akka/akka-persistence-r2dbc/pull/130/commits/acf9fdab5911d7237bda68c77e74ca228ca15ae1

johanandren (approved):
LGTM

PR #132, opened 2021-11-07
Migration of events and snapshots from any Akka Persistence plugin

patriknw:
* It has been tested with akka-persistence-jdbc as source plugin, but it should
  work with any plugin that has support for `CurrentPersistenceIdsQuery` and
  `CurrentEventsByPersistenceIdQuery`
* It can be run while the source system is still active and it can be run multiple
  times with idempotent result.

### DIFF
--- a/migration/src/main/resources/reference.conf
+++ b/migration/src/main/resources/reference.conf
@@ -1,0 +1,25 @@
+akka.persistence.r2dbc.migration {
+
+  # Akka Persistence plugin to migrate from.
+  # You must also define plugin specific configuration
+  # and application specific serializers for events and snapshots.
+  source {
+    journal-plugin-id = "jdbc-journal"
+    snapshot-plugin-id = "jdbc-snapshot-store"
+    query-plugin-id = "jdbc-read-journal"
+  }
+
+  # R2DBC Akka Persistence plugin to migrate to.
+  # You must also define akka-persistence-r2dbc specific configuration.
+  target {
+    # this must be a configuration path of akka-persistence-r2dbc
+    persistence-plugin-id = "akka.persistence.r2dbc"
+
+    # Events are stored in batches of this size.
+    batch = 10
+  }
+
+  # How many persistence ids to migrate concurrently.
+  parallelism = 10
+
+}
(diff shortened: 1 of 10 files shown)


migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
```diff
@@ -0,0 +1,324 @@
...
+    loadSourceSnapshot(persistenceId, progressSeqNr + 1).flatMap {
+      case None => Future.successful(0)
+      case Some(selectedSnapshot @ SelectedSnapshot(snapshotMetadata, _)) =>
+        for {
+          seqNr <- {
+            val serializedRow = serializedSnapotRow(selectedSnapshot)
+            targetSnapshotDao
+              .store(serializedRow)
+              .recoverWith { case _: R2dbcDataIntegrityViolationException =>
+                // ok, already exists
```
johanandren, line 270: Isn't this the upsert case for snapshots (so never any integrity violation?)

patriknw: right, I'll remove that

johanandren (approved):

migration/src/main/resources/reference.conf
```diff
@@ -0,0 +1,25 @@
+akka.persistence.r2dbc.migration {
+
+  # Akka Persistence plugin to migrate from.
+  # You must also define plugin specific configuration
+  # and application specific serializers for events and snapshots.
+  source {
+    journal-plugin-id = "jdbc-journal"
```
octonato, line 7: I think this setting is only used in tests, so maybe we can remove it to avoid confusion

octonato (approved):
LGTM

For the queries that we current support makes sense. It won't be the case if we also want to migrate tags. Then, we will need some extra DAO methods returning the tags as well. Anyway, that's for later.

migration/src/main/resources/reference.conf
patriknw: thx, removed

PR #133, opened 2021-11-08
Possibility to preserve event timestamp when writing from MigrationTool

patriknw:
need this for #132

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
@@ -146,19 +146,22 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
             .bindNull(11, classOf[Array[Byte]])
       }
 
-      val p =
-        if (useTimestampFromDb) {
-          12
-        } else {
-          stmt.bind(12, write.dbTimestamp)
-          13
-        }
+      if (useTimestampFromDb) {
+        stmt
+          .bind(12, slice)
+          .bind(13, entityType)
+          .bind(14, write.persistenceId)
+          .bind(15, previousSeqNr)
+      } else {
+        stmt
+          .bind(12, write.dbTimestamp)
+          .bind(13, slice)
+          .bind(14, entityType)
+          .bind(15, write.persistenceId)
+          .bind(16, previousSeqNr)
+      }
 
       stmt
-        .bind(p, slice)
-        .bind(p + 1, entityType)
-        .bind(p + 2, write.persistenceId)
-        .bind(p + 3, previousSeqNr)
     }
 
     val result = {
--- a/core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
@@ -74,13 +74,27 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
 
   private val journalTable = journalSettings.journalTableWithSchema
 
-  // The subselect of the db_timestamp of previous seqNr for same pid is to ensure that db_timestamp is
-  // always increasing for a pid (time not going backwards).
-  // TODO we could skip the subselect when inserting seqNr 1 as a possible optimization
-  private val insertEventSql = s"INSERT INTO $journalTable " +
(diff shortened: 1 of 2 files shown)


core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -91,15 +105,27 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
 
-  def writeEvents(events: Seq[SerializedJournalRow]): Future[Unit] = {
+  def writeEvents(events: Seq[SerializedJournalRow], samePersistenceId: Boolean): Future[Unit] = {
     require(events.nonEmpty)
-    val persistenceId = events.head.persistenceId
 
-    val entityType = SliceUtils.extractEntityTypeFromPersistenceId(persistenceId)
-    val slice = SliceUtils.sliceForPersistenceId(persistenceId, journalSettings.maxNumberOfSlices)
-    val previousSeqNr = events.head.sequenceNr - 1
+    // normally it is the same persistenceId for all events, but when used from MigrationTool it can be
```
patriknw, line 111: thinko; it's same persistenceId also from MigrationTool, so I'll revert some of this

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -125,18 +146,29 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
+
       stmt
-        .bind(12, slice)
-        .bind(13, entityType)
-        .bind(14, write.persistenceId)
-        .bind(15, previousSeqNr)
+        .bind(p, slice)
+        .bind(p + 1, entityType)
+        .bind(p + 2, write.persistenceId)
+        .bind(p + 3, previousSeqNr)
```
octonato, line 161: I think it will be easier to follow this if we just repeat the bindings inside the `if/else`.
Like we did about when matching the option in `write.metadata`.

```scala
if (useTimestampFromDb)
 stmt
    .bind(12, slice)
    .bind(13, entityType)
    .bind(14, write.persistenceId)
    .bind(15, previousSeqNr)
else
  stmt
    .bind(12, write.dbTimestamp)
    .bind(13, slice)
    .bind(14, entityType)
    .bind(15, write.persistenceId)
    .bind(16, previousSeqNr)

```

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -74,13 +74,27 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
+      s"(SELECT db_timestamp + '1 microsecond'::interval FROM $journalTable " +
+      s"WHERE slice = $$$p AND entity_type = $$${p + 1} AND persistence_id = $$${p + 2} AND sequence_number = $$${p + 3})"
+
+    val insertEventWithParameterTimestampSql =
+      baseSql + "GREATEST($13, " + insertSubSelect(p = 14) + "))"
+
+    val insertEventWithTransactionTimestampSql =
+      baseSql + "GREATEST(transaction_timestamp(), " + insertSubSelect(p = 13) + "))"
+
+    (insertEventWithParameterTimestampSql, insertEventWithTransactionTimestampSql)
```
octonato, line 96: The shift of param bindings is annoying and obfuscate a bit the code intention. 
Maybe better if we just use out-of-order params? I'm not sure if keeping the ordering is buying use anything.

octonato (approved):
LGTM

I have some comments about the param shifting, but not that important. Ok for merge as is.

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
patriknw: agreed, let me change this

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
patriknw: yeah, it's annoying. I'd prefer to have clean sql strings in the end because those may show up in logs and such.
we could add some small utility that does some string replacement of `$` to add the numbered $1, $2, $3.

patriknw: created issue https://github.com/akka/akka-persistence-r2dbc/issues/135 for this

PR #137, opened 2021-11-09
Use batch statements

### LINKED ISSUE #87: Try using batch statements again
Something didn't work when I tried them, but maybe that was related to the concurrency problems we had.

Note that there are two types of batch statements.
1) Statements with bind parameters that are separated with `.add()`
2) Batch without bind parameters

I don't think 2 is interesting for us because we want use bind parameters (e.g. binary param).

octonato:
Fixes #87

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
@@ -27,6 +27,7 @@ import org.reactivestreams.Subscriber
 import org.reactivestreams.Subscription
 import org.slf4j.Logger
 import reactor.core.publisher.Flux
+import reactor.core.publisher.Mono
 
 /**
  * INTERNAL API
@@ -57,38 +58,6 @@ import reactor.core.publisher.Flux
       promise.future
     }
 
-    def asFutureSeq(): Future[immutable.IndexedSeq[T]] =
-      asFutureSeq(identity)
-
-    def asFutureSeq[U](mapFunc: T => U): Future[immutable.IndexedSeq[U]] = {
-      val promise = Promise[immutable.IndexedSeq[U]]
-      val builder = IndexedSeq.newBuilder[U]
-
-      publisher.subscribe(new Subscriber[T] {
-
-        var subscription: Subscription = _
-
-        override def onSubscribe(s: Subscription): Unit = {
-          s.request(1)
-          subscription = s
-        }
-
-        override def onNext(value: T): Unit = {
-          builder += mapFunc(value)
-          subscription.request(1)
-        }
-
-        override def onError(t: Throwable): Unit =
-          promise.tryFailure(t)
-
-        override def onComplete(): Unit =
-          promise.trySuccess(builder.result())
-
-      })
-
-      promise.future
-    }
-
     def asFutureDone(): Future[Done] = {
       val promise = Promise[Done]()
       publisher.subscribe(new Subscriber[Any] {
@@ -113,6 +82,18 @@ import reactor.core.publisher.Flux
     }
   }
 
+  final implicit class FluxOps[T](val flux: Flux[T]) extends AnyVal {
+
+    def toIndexedSeq: Mono[immutable.IndexedSeq[T]] = toIndexedSeq(identity)
+
+    def toIndexedSeq[U](map: T => U): Mono[immutable.IndexedSeq[U]] = {
+      val consumer: BiConsumer[mutable.Builder[U, IndexedSeq[U]], T] = (builder, elem) => builder += map(elem)
+      flux
+        .collect(() => IndexedSeq.newBuilder[U], consumer)
+        .map(_.result())
+    }
+  }
+
   def updateOneInTx(stmt: Statement)(implicit ec: ExecutionContext): Future[Int] =
     stmt.execute().asFuture().flatMap { result =>
       result.getRowsUpdated.asFuture().map(_.intValue())(ExecutionContext.parasitic)
@@ -122,8 +103,9 @@ import reactor.core.publisher.Flux
     Flux
       .from[Result](stmt.execute())
       .concatMap(_.getRowsUpdated)
-      .asFutureSeq(_.intValue())
+      .toIndexedSeq(_.intValue())
       .map(_.sum)
+      .asFuture()
 
   def updateInTx(statements: immutable.IndexedSeq[Statement])(implicit
       ec: ExecutionContext): Future[immutable.IndexedSeq[Int]] =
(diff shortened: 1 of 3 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -199,6 +237,12 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger)(impli
     }
   }
 
+  def updateInBatch(logPrefix: String)(statement: Connection => Statement): Future[Int] =
+    withConnection(logPrefix) { connection =>
+      val boundStmt = statement(connection)
+      updateBatchInTx(boundStmt)
+    }
```
octonato, line 244: Needed a different method here. updateOne is not what we need because it use autoCommit and then we get a different timestamp each time. 

We need a transaction to run the batch.

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -81,15 +113,21 @@ import reactor.core.publisher.Flux
...
+  def updateOneInTx(stmt: Statement)(implicit ec: ExecutionContext): Future[Int] =
     stmt.execute().asFuture().flatMap { result =>
       result.getRowsUpdated.asFuture().map(_.intValue())(ExecutionContext.parasitic)
     }
-  }
+
+  def updateBatchInTx(stmt: Statement)(implicit ec: ExecutionContext): Future[Int] =
+    Flux
+      .from[Result](stmt.execute())
+      .concatMap(_.getRowsUpdated)
```
octonato, line 124: concatMap is what allow us to force the execution of all inserts.

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -57,6 +57,38 @@ import reactor.core.publisher.Flux
       promise.future
     }
 
+    def asFutureSeq(): Future[immutable.IndexedSeq[T]] =
+      asFutureSeq(identity)
+
+    def asFutureSeq[U](mapFunc: T => U): Future[immutable.IndexedSeq[U]] = {
```
patriknw, line 63: This is becoming too much "implement your own reactive streams ...". The reason for doing that for the simple Future and Future[Done] cases are that those are used for very fine grained operations and I want them to be as fast as possible.

Here we should use the reactor api or Akka Streams `Source.fromPublisher`.

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -199,6 +237,12 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger)(impli
     }
   }
 
+  def updateInBatch(logPrefix: String)(statement: Connection => Statement): Future[Int] =
+    withConnection(logPrefix) { connection =>
```
patriknw, line 241: Good, as far as I can see we can change the `update` method to also use `withConnection`. The only difference would be the debug logging of number of rows updated, but that is not important. Could you change that at the same?

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
patriknw: Probably best to use the reactor api since we already use that in updateBatchInTx

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -165,32 +165,21 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
-      } else {
-        // TODO batch statements doesn't work, see above
-        r2dbcExecutor
-          .update(s"insert [$persistenceId]") { connection =>
-            events.map { write =>
-              val stmt = connection.createStatement(insertSql)
-              bind(stmt, write)
-            }.toIndexedSeq
+      else
+        r2dbcExecutor.updateInBatch(s"batch insert [$persistenceId] ") { connection =>
```
patriknw, line 178:
```suggestion
        r2dbcExecutor.updateInBatch(s"batch insert [$persistenceId], [${events.size}] events") { connection =>
```

patriknw (commented):
looking good

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
octonato: yes, that's what I was planning to change on top of this PR, but didn't want to add too much here. 

Also I think could have two methods. The `withConnection` is more about a connection with a open transaction. 
So maybe we should have `withConnection` using auto-comit and `withConnectionTx` for the transactional one. Or some other naming.

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
octonato: I didn't want to use Source.fromPublisher because we had issues with iterations with the driver. Although I think here is different. The issue we had was in updating. 

I did a first impl using reactor API, but changed to this. The reason I changed is that I was planning to use the same in the updateBatchInTx. I want to have one single way of taking a Flux and building a Scala IndexedSeq out of it without repeating what we did in updateBatchInTx.

Anyway, I think we add a method to a Flux that does that without using subscription. I will try this out.

octonato (commented):
Code suggestions applied. Ready for one more round.

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -81,15 +82,34 @@ import reactor.core.publisher.Flux
...
     }
-  }
+
+  def updateBatchInTx(stmt: Statement)(implicit ec: ExecutionContext): Future[Int] =
+    Flux
+      .from[Result](stmt.execute())
+      .concatMap(_.getRowsUpdated)
+      .toIndexedSeq(_.intValue())
+      .map(_.sum)
+      .asFuture()
```
johanandren, line 108: Not super important perhaps, but, could we collect/fold into sum to start with instead of into a seq and then sum?

johanandren (approved):

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
octonato: yes, we can. I will do it. 

Then we can also remove the FluxOps as it will be come obsolete

octonato: Here 7da2f41

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -427,7 +417,7 @@ private[projection] class R2dbcOffsetStore(
         } else
           newState
 
-      R2dbcExecutor.updateInTx(stmts).map { _ =>
+      R2dbcExecutor.updateBatchInTx(batchStatement).map { _ =>
```
patriknw, line 420: small optimization could be to use `updateOneInTx` when there is only one record (which is the normal case for exactly-once)

patriknw (approved):
LGTM, with a small performance suggestion

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
octonato: like this 74c1f26 ?

octonato:
> just a typo. "bounded" is a different thing.
What we talk about here is the irregular verb: bind / bound / bound

Yeah, correct. Bound as in bind, not as in bounded / limited :-)

PR #138, opened 2021-11-10
Lazy loading of events/state for backtracking queries, #76

### LINKED ISSUE #76: Optimize backtracking data fetch
Assuming that missed events are rare it's wasteful to fetch the full event payload in the backtracking queries.

One simple optimization would be to just do the deserialization work lazyily when verified that the event will be used.

Another approach would be to not fetch the event payload at all in the backtracking queries and instead fetch that for individual pid/seqNr from the journal table when when verified that the event will be used.

patriknw:
Fixes #76

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
@@ -43,6 +43,7 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
     with CurrentEventsBySliceQuery
     with EventsBySliceQuery
     with EventTimestampQuery
+    with LoadEventQuery
     with CurrentEventsByPersistenceIdQuery
     with EventsByPersistenceIdQuery
     with CurrentPersistenceIdsQuery {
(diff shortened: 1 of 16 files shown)


core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -99,6 +100,34 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+   * `transaction_timestamp()` is the time when the transaction started, not when it was committed. This means that a
+   * "later" event may be visible first and when retrieving events after the previously seen timestamp may miss some
+   * events. In distributed SQL databases there can also be clock skews for the database timestamps. For that reason it
+   * will perform additional backtracking queries to catch missed events. Events from backtracking will typically be
+   * duplicates of previously emitted events. It's the responsibility of the consumer to filter duplicates and make sure
+   * that events are processed in exact sequence number order for each persistence id. Such deduplication is provided by
+   * the R2DBC Projection.
+   *
+   * Events emitted by the backtracking don't contain the event payload (`EventEnvelope.event` is null) and the consumer
+   * can load the full `EventEnvelope` with [[R2dbcReadJournal.loadEnvelope]].
```
patriknw, line 122: It's ofc undesired to use `null` for the payload. I'm thinking about introducing a new EventEnvelope which can support that better. Maybe something more specific to the new eventsBySlices query. We have anyway some mess with two EventEnvelopes, one in akka.persistence.query and another in akka.projection

patriknw: but that would be another PR since it would be a very intrusive change

ddl-scripts/create_tables_yugabyte.sql
```diff
@@ -20,7 +20,8 @@ CREATE TABLE IF NOT EXISTS event_journal(
   PRIMARY KEY((slice, entity_type) HASH, persistence_id, seq_nr ASC)
 );
 
-CREATE INDEX IF NOT EXISTS event_journal_slice_idx ON event_journal(entity_type ASC, slice ASC, db_timestamp ASC);
+CREATE INDEX IF NOT EXISTS event_journal_slice_idx ON event_journal(entity_type ASC, slice ASC, db_timestamp ASC)
+  INCLUDE (seq_nr, deleted);
```
johanandren, line 24: So `seq_nr` and `deleted` goes in the index but without being keys of the index or what does `INCLUDE` do?

johanandren (approved):
LGTM (but messy with the null)

ddl-scripts/create_tables_yugabyte.sql
patriknw: Yes, good that you spotted that, I should have mentioned it. That will be much cheaper because then the backtracking queries don't have to do the join (lookup) with the parent table (possibly on a remote tablet). It will only use the index.

octonato: Is this an specific Yugabyte thing? 
That's quite interesting actually.

octonato (approved):
LGTM, but indeed the `null` feels hacky, but that's how it is. 

Agree we can improved later. It's true that the persistence APIs in general are quite old and since their inception we have a few confirmed requirements/use cases (projection, slicing, backtracking) requiring us to re-think it a bit.

ddl-scripts/create_tables_yugabyte.sql
patriknw: yup: https://docs.yugabyte.com/latest/api/ysql/the-sql-language/statements/ddl_create_index/#include-clause

It's a way to create a consistent materialized view via an index.

One could include all columns in the index, at the cost of larger size on disk (and maybe slightly slower writes). Nice thing is that a user of this plugin could change the index definition if that is preferred and all queries would still be the same.

patriknw:
> but indeed the null feels hacky, but that's how it is

I have started the introduction of a new `EventBySliceEnvelope`. I don't think there is good reason for reusing the old EventEnvelope for this new query and projection.

PR #141, opened 2021-11-10
minor improments on withConnection method usage

octonato:
Draft because on top of #137.

Ready for review otherwise.

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
@@ -5,7 +5,6 @@
 package akka.persistence.r2dbc.internal
 
 import java.util.function.BiConsumer
-import java.util.function.Supplier
 
 import scala.collection.immutable
 import scala.collection.mutable
@@ -20,6 +19,7 @@ import akka.actor.typed.ActorSystem
 import akka.annotation.InternalStableApi
 import io.r2dbc.spi.Connection
 import io.r2dbc.spi.ConnectionFactory
+import io.r2dbc.spi.Result
 import io.r2dbc.spi.Row
 import io.r2dbc.spi.Statement
 import org.reactivestreams.Publisher
@@ -27,6 +27,7 @@ import org.reactivestreams.Subscriber
 import org.reactivestreams.Subscription
 import org.slf4j.Logger
 import reactor.core.publisher.Flux
+import reactor.core.publisher.Mono
 
 /**
  * INTERNAL API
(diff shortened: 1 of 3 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -138,74 +138,21 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger)(impli
...
-        rowsUpdated.failed.foreach { exc =>
-          log.debug("{} - Update failed: {}", logPrefix, exc)
-          // auto-commit so nothing to rollback
-          connection.close().asFutureDone()
-        }
-
-        rowsUpdated.flatMap { r =>
-          connection.close().asFutureDone().map { _ =>
-            if (log.isDebugEnabled())
-              log.debug("{} - Updated [{}] in [{}] µs", logPrefix, r, (System.nanoTime() - startTime) / 1000)
```
octonato, line 202: There is not much different in terms of logging, so reusing new `withAutoCommitConnection`

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -138,74 +138,21 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger)(impli
...
-            r
-          }
-        }
-      }
+  def updateOne(logPrefix: String)(statementFactory: Connection => Statement): Future[Int] =
+    withAutoCommitConnection(logPrefix) { connection =>
```
octonato, line 153: we could eventually add `logPrefix + "(UpdateOne)"` just to add more context.

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -138,74 +138,21 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger)(impli
...
-
-        done.flatMap { _ =>
-          connection.close().asFutureDone().map { _ =>
-            if (log.isDebugEnabled())
-              log.debug("{} - DDL in [{}] µs", logPrefix, (System.nanoTime() - startTime) / 1000)
-            Done
-          }
-        }
+  def executeDdl(logPrefix: String)(statement: Connection => Statement): Future[Done] =
+    withAutoCommitConnection(logPrefix + " (DDL)") { connection =>
```
octonato, line 142: Enriched prefix for extra context. Maybe something we should do each time.

octonato:
Actually, I'm not so sure anymore if we should append even more information on the logPrefix like I did here with the DDL. 
At the end, the original log prefix should contain enough info. If not, then we are not using is as it should. 

Looking for other opinions.

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
octonato: I removed this extra prefix info. Not convinced is a good idea.

patriknw:
> Actually, I'm not so sure anymore if we should append even more information on the logPrefix like I did here with the DDL.
At the end, the original log prefix should contain enough info. If not, then we are not using is as it should.

I think I agree that it's better to let the caller define the full information.

patriknw (approved):
LGTM, great cleanup

lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

octonato:
no clear what happened here. Something closed this PR and now Lightbend CLA validator is not even recognizing me anymore. 🤦 

Anyway, no time for this now. Tomorrow I check it again.

Issue #135, opened 2021-11-08
Utility to replace numbered sql parameters


patriknw:
To improve the code readability we can implement a small string search and replace utility that replaces `$` with numbered `$1`, `$2`, `$3`.

johanandren:
Is it up to the r2dbc plugin to support named parameter to start with and the postgres one doesn't (if so, could we contribute that perhaps)?

patriknw:
In the end I want to bind with index because that is less lookups when binding parameters. Probably a very minor performance difference, but anyway.

PR #142, opened 2021-11-10
pause/resume projection management

octonato:
References #100

### DIFF
--- a/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -34,6 +34,7 @@ import akka.projection.internal.OffsetSerialization.MultipleOffsets
 import akka.projection.internal.OffsetSerialization.SingleOffset
 import akka.projection.r2dbc.R2dbcProjectionSettings
 import io.r2dbc.spi.Connection
+import io.r2dbc.spi.Row
 import io.r2dbc.spi.Statement
 import org.slf4j.LoggerFactory
 
@@ -151,6 +152,7 @@ private[projection] class R2dbcOffsetStore(
 
   private val timestampOffsetTable = settings.timestampOffsetTableWithSchema
   private val offsetTable = settings.offsetTableWithSchema
+  private val managementTable = settings.managementTableWithSchema
 
   // FIXME define this in akka.persistence.Persistence (not per plugin)
   private val maxNumberOfSlices =
@@ -194,6 +196,44 @@ private[projection] class R2dbcOffsetStore(
 
   private val clearOffsetSql: String =
     s"""DELETE FROM $offsetTable WHERE projection_name = $$1 AND projection_key = $$2"""
+
+  object ReadManagementStateQuery {
+
+    val projectionName: Int = 0
+    val projectionKey: Int = 1
+
+    private def bindParam(i: Int) = "$" + (i + 1)
+
+    val statement: String =
+      s"SELECT paused FROM $managementTable WHERE " +
+      s"projection_name = ${bindParam(projectionName)} and " +
+      s"projection_key = ${bindParam(projectionKey)} "
+
+    /* Map rows retrieved by this query */
+    def mapRow(row: Row) =
+      ManagementState(row.get("paused", classOf[java.lang.Boolean]))
+  }
+
+  object UpdateManagementStateQuery {
+
+    val projectionName: Int = 0
+    val projectionKey: Int = 1
+    val pausedParam: Int = 2
+    val lastUpdate: Int = 3
+
+    private def bindParam(i: Int) = "$" + (i + 1)
+
+    val statement: String = {
+      s"INSERT INTO $managementTable " +
+      "(projection_name, projection_key, paused, last_updated)  " +
+      "VALUES " +
+      s"(${bindParam(projectionName)},${bindParam(projectionKey)}, ${bindParam(pausedParam)}, ${bindParam(lastUpdate)}) " +
+      "ON CONFLICT (projection_name, projection_key)" +
+      "DO UPDATE SET " +
+      "paused = excluded.paused, " +
+      "last_updated = excluded.last_updated"
+    }
+  }
 
   // The OffsetStore instance is used by a single projectionId and there shouldn't be any concurrent
   // calls to methods that access the `state`. To detect any violations of that concurrency assumption
@@ -685,11 +725,36 @@ private[projection] class R2dbcOffsetStore(
   }
 
   def readManagementState(): Future[Option[ManagementState]] = {
-    Future.successful(None) // FIXME not implemented yet
+    import ReadManagementStateQuery._
+
+    def createStatement(connection: Connection) =
+      connection
+        .createStatement(statement)
+        .bind(projectionName, projectionId.name)
+        .bind(projectionKey, projectionId.key)
+
+    r2dbcExecutor
+      .selectOne("read management state")(conn => createStatement(conn), row => mapRow(row))
   }
 
   def savePaused(paused: Boolean): Future[Done] = {
-    Future.successful(Done) // FIXME not implemented yet
+    import UpdateManagementStateQuery._
+
+    r2dbcExecutor
+      .updateOne("update management state") { conn =>
+        conn
+          .createStatement(statement)
+          .bind(projectionName, projectionId.name)
+          .bind(projectionKey, projectionId.key)
+          .bind(pausedParam, paused)
+          .bind(lastUpdate, Instant.now(clock).toEpochMilli)
+      }
+      .flatMap {
+        case i if i == 1 => Future.successful(Done)
+        case n =>
+          Future.failed(
+            new RuntimeException(s"Failed to update management table for ${projectionId.name} and ${projectionId.key}"))
+      }
   }
 
   private def createRecordWithOffset[Envelope](envelope: Envelope): Option[RecordWithOffset] = {
--- a/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -752,8 +752,7 @@ private[projection] class R2dbcOffsetStore(
       .flatMap {
         case i if i == 1 => Future.successful(Done)
         case n =>
-          Future.failed(
-            new RuntimeException(s"Failed to update management table for $projectionId"))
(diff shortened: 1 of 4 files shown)


projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -195,6 +197,44 @@ private[projection] class R2dbcOffsetStore(
...
+
+    val statement: String =
+      s"SELECT paused FROM $managementTable WHERE " +
+      s"projection_name = ${bindParam(projectionName)} and " +
+      s"projection_key = ${bindParam(projectionKey)} "
+
+    /* Map rows retrieved by this query */
+    def mapRow(row: Row) =
+      ManagementState(row.get("paused", classOf[java.lang.Boolean]))
+  }
```
octonato, line 215: By wrapping the statement and binding logic in such an object, I think we can make it easier to deal with the bindings. 

I'm thinking on the statements that has variable bindings. We can encapsulate the numbers and eventual index shifting in such a query object.

johanandren: I find this harder to follow than the original indexes.

octonato: Are you referring to usage of names like `projectionName` instead of int indices? 
Like `bind(0, "projection name")` is better than `bind(projectName, "projection name")`? 

Or the fact that this is encapsulating the whole query binding requirements in one single place? 

There is one thing that this completes remove and that I find hard to keep track is the fact the index `0` refers to `$1`. Maybe not a big deal for small queries, but when we start to have variable params, index shifting or things like `nextParam()` we are just making the code harder to ourselves.

My main goal here is to experiment with something that could solve #135 without resorting to extra string manipulations (find/replace). Keeping the indices, but in a way that we refer to them by name. 

And of course, each query is different and such an object/class could then encapsulate the specifics of each query (like param shifting and alike).

patriknw: I think there is a misunderstanding. I think it's just a convention (maybe some special parsing in r2dbc-postgres) for using parameter names "$1", "$2", "$3". I don't think those numbers are related to the index numbers 0, 1, 2.

That's why I think we should write the sql without numbers.
```
select * from Foo where a = $ and b = $
```
and then have the simple search-replace utility to construct
```
select * from Foo where a = $1 and b = $2
```

johanandren: I meant the indirection to some the query index constant for simple queries vs just using them in string.

Could we try adding support for named params in prepared statements upstream instead (since R2DBC has API for it) or is that completely up to the database flavour rather than the driver?

johanandren: For the record I just expected I could use that when starting out: 

```
.createStatement("SELECT yadi FROM TABLE WHERE yadi > $theParam")
    .bind("theParam", 42)
```

patriknw: > encapsulating the whole query binding requirements in one single place

When we add support for other databases I think it's these pieces that should be pluggable, not the full Dao classes.
* sql statement
* bind parameters
* mapRow

However, it's a bit early to make that refactoring. Should be done together with adding next database.

patriknw: > Could we try adding support for named params in prepared statements upstream instead (since R2DBC has API for it) or is that completely up to the database flavour rather than the driver?

I'm not sure but I think that is already supported. At least it works with named parameters as `bind("$1", ...)`. That was what I started with but changed to index in #21 because it's a bit more efficient (no name lookups).

projection/src/test/scala/akka/projection/r2dbc/R2dbcProjectionSpec.scala
```diff
@@ -1295,7 +1295,6 @@ class R2dbcProjectionSpec
     }
 
     "pause projection" in {
-      pending // FIXME not implemented yet
```
patriknw, line 1298: there are a few more pending in R2dbcOffsetStoreSpec, R2dbcTimestampOffsetStoreSpec, R2dbcTimestampOffsetProjectionSpec

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -685,11 +725,36 @@ private[projection] class R2dbcOffsetStore(
...
+          .bind(projectionName, projectionId.name)
+          .bind(projectionKey, projectionId.key)
+          .bind(pausedParam, paused)
+          .bind(lastUpdate, Instant.now(clock).toEpochMilli)
+      }
+      .flatMap {
+        case i if i == 1 => Future.successful(Done)
+        case n =>
+          Future.failed(
+            new RuntimeException(s"Failed to update management table for ${projectionId.name} and ${projectionId.key}"))
```
patriknw, line 756:
```suggestion
            new RuntimeException(s"Failed to update management table for $projectionId"))
```

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
johanandren: I see, I thought it simply didn't work, but I probably missed out on the $ in the bind param name.

octonato: > is that completely up to the database flavour rather than the driver 

They explicit say in their docs that it's not the goal to solve that in the driver. 

_The SQL that is used to create a statement can be parameterized by using vendor-specific bind markers. The portability of SQL statements across R2DBC implementations is not a goal._

I also thought that using $something and then passing "something" to the bind method would work out of the box. I'm quite sure that I also passed '$something' without success. Anyway, I will try again.

> When we add support for other databases I think it's these pieces that should be pluggable

I had that in mind as well, but it's too early for it. 
At some point, extracting the queries and the binding logic will become mandatory.

octonato: Hmm, I went back to see how we did in projections and we have:

* placeholders using `?` for all dialects
* objects indices so we could easily link a value to some name, instead of a meanless int, eg: `UpdateManagementIndices.PAUSED` 
* surprisingly the indices use base 1

I'm wondering if that's more a R2DBC issue and not really DB specific.

octonato: > However, it's a bit early to make that refactoring. Should be done together with adding next database. 

I missed that comment. So, yes, I also think it's early. I said on my previous comment. 

My idea was more related to make the binding easy to follow. SInce it's a very simple query, I thought it was a good place to experiment with something without having to touch much code. 

But true that while doing it it became clear to me that we will need to abstract the queries and binding away because of different dialects.

projection/src/test/scala/akka/projection/r2dbc/R2dbcProjectionSpec.scala
octonato: `pending` removed, but not in `R2dbcTimestampOffsetProjectionSpec`. 

There we have another related with how we read the current offset. We don't take fill in the `readTimestamp` and it's always EPOCH causing the test to fail. I think we should fix this in another PR.

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -195,6 +197,21 @@ private[projection] class R2dbcOffsetStore(
...
+
+  val updateManagementStateSql: String =
+    s"INSERT INTO $managementTable " +
+    "(projection_name, projection_key, paused, last_updated)  " +
+    "VALUES " +
+    "($1,$2,$3,$4) " +
+    "ON CONFLICT (projection_name, projection_key) " +
+    "DO UPDATE SET " +
+    "paused = excluded.paused, " +
+    "last_updated = excluded.last_updated"
```
octonato, line 213: I remove the query objects.

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -685,11 +702,33 @@ private[projection] class R2dbcOffsetStore(
...
   def savePaused(paused: Boolean): Future[Done] = {
-    Future.successful(Done) // FIXME not implemented yet
+    r2dbcExecutor
+      .updateOne("update management state") { conn =>
+        conn
+          .createStatement(updateManagementStateSql)
+          .bind("$1", projectionId.name)
+          .bind("$2", projectionId.key)
+          .bind("$3", paused)
+          .bind("$4", Instant.now(clock).toEpochMilli)
```
octonato, line 725: I tried a few options and the only thing that seems to work is: 

* "$1" with `bind(0, ...)` 
* "$1" with `bind("$1", ...)` 

Using a name like `$projection_name` doesn't work. 

I still don't understand why we prefer to use `bind(0, ...)` or `bind("$1", ...)`, instead of constants with clear names.

patriknw: ok, I only tried `bind("$1", ...)` and assumed it would work for any string.

> I still don't understand why we prefer to use bind(0, ...), instead of constants with clear names.

I don't mind adding constants for the indexes in the bind calls.

For the sql construction I don't think we should use those constants to create the "$1" in the sql string. Instead I still think the search-replace in #135 would be nice.

patriknw: we should not use `.bind("$1"`
that has no value over `.bind(0` and only adds runtime overhead

octonato: I will replace with `bind(int, ...)`

> For the sql construction I don't think we should use those constants to create the "$1"

My initial idea was to use an constant `Int`, but use it to generate the right $ param and then refer to the constant as usage sight. 

After all the discussion here, it's not clear how you want to implement #135. It seems to me that we will still need to count the order of params to know if we are using the write index. Anyway, maybe better to discuss this further in #135.

patriknw (approved):
LGTM

PR #144, opened 2021-11-11
Add new EventBySliceEnvelope

patriknw:
* that has Option event for the lazy loading case (envelopes from backtracking), see #138 
* same envelope in akka.persistence.query and akka.projection

### DIFF
--- a/core/src/main/scala/akka/persistence/query/EventBySliceEnvelope.scala
+++ b/core/src/main/scala/akka/persistence/query/EventBySliceEnvelope.scala
@@ -1,0 +1,104 @@
+/*
+ * Copyright (C) 2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.query
+
+import java.util.Optional
+
+import akka.util.HashCode
+
+object EventBySliceEnvelope {
+  def apply[Event](
+      offset: Offset,
+      persistenceId: String,
+      sequenceNr: Long,
+      event: Event,
+      timestamp: Long,
+      entityType: String,
+      slice: Int): EventBySliceEnvelope[Event] =
+    new EventBySliceEnvelope(offset, persistenceId, sequenceNr, Option(event), timestamp, None, entityType, slice)
+
+  def create[Event](
+      offset: Offset,
+      persistenceId: String,
+      sequenceNr: Long,
+      event: Event,
+      timestamp: Long,
+      entityType: String,
+      slice: Int): EventBySliceEnvelope[Event] =
+    apply(offset, persistenceId, sequenceNr, event, timestamp, entityType, slice)
+
+  def unapply[Event](arg: EventBySliceEnvelope[Event]): Option[(Offset, String, Long, Option[Event], Long)] =
+    Some((arg.offset, arg.persistenceId, arg.sequenceNr, arg.eventOption, arg.timestamp))
+}
+
+/**
+ * Event wrapper adding meta data for the events in the result stream of
+ * [[akka.persistence.query.scaladsl.EventsBySliceQuery]] query, or similar queries.
+ *
+ * If the `event` is not defined it has not been loaded yet. It can be loaded with the
+ * [[akka.persistence.query.scaladsl.LoadEventQuery]].
+ *
+ * The `timestamp` is the time the event was stored, in milliseconds since midnight, January 1, 1970 UTC (same as
+ * `System.currentTimeMillis`).
+ */
+final class EventBySliceEnvelope[Event](
+    val offset: Offset,
+    val persistenceId: String,
+    val sequenceNr: Long,
+    val eventOption: Option[Event],
+    val timestamp: Long,
+    val eventMetadata: Option[Any],
+    val entityType: String,
+    val slice: Int) {
+
+  def event: Event =
+    eventOption match {
+      case Some(evt) => evt
+      case None =>
+        throw new IllegalStateException(
+          "Event was not loaded. Use getOptionalEvent and load the event on demand with LoadEventQuery.")
+    }
+
+  /**
+   * Java API
+   */
+  def getEvent(): Event =
+    event
+
+  /**
+   * Java API
+   */
+  def getOptionalEvent(): Optional[Event] = {
+    import scala.compat.java8.OptionConverters._
+    eventOption.asJava
+  }
+
+  /**
+   * Java API
+   */
+  def getEventMetaData(): Optional[AnyRef] = {
+    import scala.compat.java8.OptionConverters._
+    eventMetadata.map(_.asInstanceOf[AnyRef]).asJava
+  }
+
+  override def hashCode(): Int = {
+    var result = HashCode.SEED
+    result = HashCode.hash(result, offset)
+    result = HashCode.hash(result, persistenceId)
+    result = HashCode.hash(result, sequenceNr)
+    result
+  }
+
+  override def equals(obj: Any): Boolean = obj match {
+    case other: EventBySliceEnvelope[_] =>
+      offset == other.offset && persistenceId == other.persistenceId && sequenceNr == other.sequenceNr &&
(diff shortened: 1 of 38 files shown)


core/src/main/scala/akka/persistence/query/scaladsl/EventsBySliceQuery.scala
```diff
@@ -19,7 +19,22 @@ import akka.stream.scaladsl.Source
...
+  //   query.eventsBySlices(...)
+  // one would have to cast it to more specific EventsBySliceQuery[TheEvent]
+  // or use that in the readJournalFor class
+  // on the other hand, the alternative is also defining the type:
+  // query.eventsBySlices[TheEvent](...)
+  //
+  // Maybe it should be on the def because the same R2dbcReadJournal instance can be used for different entityTypes,
+  // i.e. different Event classes
+  //
+  // The original reason I changed it was some trouble with similar param in LoadEventQuery
```
patriknw, line 37: I'm on the fence here and would like input/opinions

core/src/main/scala/akka/persistence/query/EventBySliceEnvelope.scala
```diff
@@ -0,0 +1,104 @@
...
+ * Event wrapper adding meta data for the events in the result stream of
+ * [[akka.persistence.query.scaladsl.EventsBySliceQuery]] query, or similar queries.
+ *
+ * If the `event` is not defined it has not been loaded yet. It can be loaded with the
+ * [[akka.persistence.query.scaladsl.LoadEventQuery]].
+ *
+ * The `timestamp` is the time the event was stored, in milliseconds since midnight, January 1, 1970 UTC (same as
+ * `System.currentTimeMillis`).
+ */
+final class EventBySliceEnvelope[Event](
```
patriknw, line 46: replaces `akka.persistence.query.EventEnvelope` and `akka.projection.eventsourced.EventEnvelope` for eventsBySlices

core/src/main/scala/akka/persistence/query/scaladsl/EventsBySliceQuery.scala
octonato: I think it should be on the method for the reasons you mentioned. We can store different events types.

I don't get which kind of trouble you had with `LoadEventQuery`. Could you clarify?

core/src/main/scala/akka/persistence/query/EventBySliceEnvelope.scala
```diff
@@ -0,0 +1,104 @@
...
+ * Copyright (C) 2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.query
+
+import java.util.Optional
+
+import akka.util.HashCode
+
+object EventBySliceEnvelope {
```
octonato, line 11: Do we need to add Slice in the name? 

In an ideal worlds, we should have one kind of envelope that is typed and supporting lazy loading. 
If we have: `akka.persistence.query.typed.EventEnvelope` we could gradually move all queries to it (with proper deprecation cycle) and also AkkaProjection. 

We don't need to impose us any deadline about it, but we can leave the door open by using a more generic name.

patriknw: > Do we need to add Slice in the name?

Yeah, I had to come up with a different name, but I can use a different package name as you suggest.

Then we should place the new query interfaces in that `typed` package as well.

core/src/main/scala/akka/persistence/query/scaladsl/EventsBySliceQuery.scala
patriknw: Thanks, I'll try with that again, first with LoadEventQuery.

core/src/main/scala/akka/persistence/query/EventBySliceEnvelope.scala
```diff
@@ -0,0 +1,104 @@
...
+ * [[akka.persistence.query.scaladsl.LoadEventQuery]].
+ *
+ * The `timestamp` is the time the event was stored, in milliseconds since midnight, January 1, 1970 UTC (same as
+ * `System.currentTimeMillis`).
+ */
+final class EventBySliceEnvelope[Event](
+    val offset: Offset,
+    val persistenceId: String,
+    val sequenceNr: Long,
+    val eventOption: Option[Event],
```
octonato, line 50: Would be benefit of making this a private `OptionVal` and only wrap it in `Option` or `Optional` on user request?

core/src/main/scala/akka/persistence/query/EventBySliceEnvelope.scala
octonato: good idea to also put the queries under `typed`. That gives us a migration and deprecation path for the existing queries.

core/src/main/scala/akka/persistence/query/EventBySliceEnvelope.scala
patriknw: Good suggestion, but I don't think it is worth it. OptionVal is internal api (InternalStableApi) so we should not encourage usage of it outside of Akka core. Would require additional overloaded constructors (factory methods) and if conditions in plugins to select constructor. E.g. SerializedJournalRow also has an Option field for the event payload.

These queries are anyway allocating rather many objects so saving one allocation here is not worth much.

core/src/main/scala/akka/persistence/query/scaladsl/EventsBySliceQuery.scala
patriknw: done in https://github.com/akka/akka-persistence-r2dbc/pull/144/commits/0b3e9ea4030ab29a38c7e96356f876c2a1cecf2d

core/src/main/scala/akka/persistence/query/EventBySliceEnvelope.scala
patriknw: done in https://github.com/akka/akka-persistence-r2dbc/pull/144/commits/f900a16926c74b682edabfd58c8490b978b58c38

core/src/main/scala/akka/persistence/r2dbc/query/javadsl/R2dbcReadJournal.scala
```diff
@@ -54,15 +68,23 @@ final class R2dbcReadJournal(delegate: scaladsl.R2dbcReadJournal)
...
       fromSequenceNr: Long,
-      toSequenceNr: Long): Source[EventEnvelope, NotUsed] =
+      toSequenceNr: Long): Source[ClassicEventEnvelope, NotUsed] =
     delegate.currentEventsByPersistenceId(persistenceId, fromSequenceNr, toSequenceNr).asJava
 
   override def eventsByPersistenceId(
       persistenceId: String,
       fromSequenceNr: Long,
-      toSequenceNr: Long): Source[EventEnvelope, NotUsed] =
+      toSequenceNr: Long): Source[ClassicEventEnvelope, NotUsed] =
```
octonato, line 77: Since we are taking the path of having `typed.EventEnvelope` we could also make this plugin only support the new typed queries. 
Anyway, just an idea for later. It will require us to bring all the typed queries into akka/akka first.

octonato (approved):
LGTM

core/src/main/scala/akka/persistence/r2dbc/query/javadsl/R2dbcReadJournal.scala
patriknw: It's mostly `eventsByPersistenceId` and `currentEventsByPersistenceId` we are talking about, right? I think we still want to have some of the "old" to be able to abstract over the plugin. For example to use `currentEventsByPersistenceId` when switching between Spanner and R2DBC. Another example is that ReplicatedEventSourcing is using `eventsByPersistenceId`.

Issue #118, opened 2021-11-04
Failed: EventsBySliceSpec


johanandren:
Against Yugabyte

```
 - should retrieve from several slices *** FAILED *** (12 seconds, 948 milliseconds)
[info]   A timeout occurred waiting for a future to complete. Waited 10000000000 nanoseconds. (EventsBySliceSpec.scala:338)
[info]   org.scalatest.concurrent.ScalaFutures$$anon$1$$anon$2:
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:339)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:281)
[info]   at akka.persistence.r2dbc.query.EventsBySliceSpec$$anon$8.<init>(EventsBySliceSpec.scala:338)
[info]   at akka.persistence.r2dbc.query.EventsBySliceSpec.$anonfun$new$25(EventsBySliceSpec.scala:296)
```
[logs_220.zip](https://github.com/akka/akka-persistence-r2dbc/files/7475843/logs_220.zip)

patriknw:
https://github.com/akka/akka-persistence-r2dbc/runs/4106845087?check_suite_focus=true#step:7:2534

patriknw:
not sure it has been fixed, but closing until seen again

PR #145, opened 2021-11-12
Harden EventsBySliceSpec, #118

patriknw:
* flaky in CI
* I looked at logs from one of the failed runs and it was missing 1 event for the
  "Live eventsBySlices should retrieve from several slices", couldn't see
  a reason from the logs

References #118


core/src/test/scala/akka/persistence/r2dbc/query/EventsBySliceSpec.scala
```diff
@@ -47,6 +45,9 @@ object EventsBySliceSpec {
   def config: Config =
     TestConfig.backtrackingDisabledConfig
       .withFallback(ConfigFactory.parseString(s"""
+    # This test is not using backtracking, so increase behind-current-time to
+    # reduce risk of missing events
+    akka.persistence.r2dbc.query.behind-current-time = 500 millis
```
patriknw, line 50: this is the attempt to harden it, but I don't have no evidence that it is the reason for the failure

PR #148, opened 2021-11-15
Rewrite test for grouped async, #147

### LINKED ISSUE #147: failed: filter out unknown sequence numbers for grouped async projection
It's marked as pending.

I found what is wrong. It's because we use `projectionTestKit.run`.  Second run will create a new offset store which ofc doesn't contain the inflight from the previous run.

patriknw:
* problem was that projectionTestKit.run resets the OffsetStore for each
  run and therefore the inflight was not kept
* use a new TestSourceProvider that can be feeded with
  envelopes rather than just emitting a fixed collection

I might rewrite some of the other tests that could have similar async problem (even though we might not have seen it yet).

Fixes #147


johanandren (approved):

octonato (approved):

patriknw:
I'll follow up and rewrite some of the other tests that could have similar async problem (even though we might not have seen it yet).

PR #149, opened 2021-11-16
Reference documentation, #89

patriknw:
Doesn't include docs for Projections yet. I'll add that in a separate PR.

References #89

### DIFF
--- a/docs/src/main/paradox/durable-state-store.md
+++ b/docs/src/main/paradox/durable-state-store.md
@@ -1,0 +1,33 @@
+# Durable state store plugin
+
+The durable state plugin enables storing and loading key-value entries for @extref:[durable state actors](akka:typed/durable-state/persistence.html).
+
+## Schema
+
+The `durable_state` table and `durable_state_slice_idx` index need to be created in the configured database, see schema definition in @ref:[Creating the schema](getting-started.md#schema).
+
+The `durable_state_slice_idx` is only needed if the slice based @ref:[queries](query.md) are used.
+
+## Configuration
+
+To enable the journal plugin to be used by default, add the following line to your Akka `application.conf`:
+
+```
+akka.persistence.state.plugin = "akka.persistence.r2dbc.durable-state-store"
+```
+
+It can also be enabled with the `durableStateStorePluginId` for a specific `DurableStateBehavior` and multiple
+plugin configurations are supported.
+
+See also @ref:[Connection configuration](connection-config.md).
+
+### Reference configuration
+
+The following can be overridden in your `application.conf` for the journal specific settings:
+
+@@snip [reference.conf](/core/src/main/resources/reference.conf) {#durable-state-settings}
+
+### Deletes
+
+The journal supports deletes through hard deletes, which means the durable state store entries are actually deleted from the database.
+There is no materialized view with a copy of the state so make sure to not delete durable states too early if they are used from projections or queries.
--- a/docs/src/main/paradox/durable-state-store.md
+++ b/docs/src/main/paradox/durable-state-store.md
@@ -6,7 +6,7 @@ The durable state plugin enables storing and loading key-value entries for @extr
 
 The `durable_state` table and `durable_state_slice_idx` index need to be created in the configured database, see schema definition in @ref:[Creating the schema](getting-started.md#schema).
 
-The `durable_state_slice_idx` is only needed if the slice based @ref:[queries](query.md) are used.
+The `durable_state_slice_idx` index is only needed if the slice based @ref:[queries](query.md) are used.
 
 ## Configuration
 
--- a/docs/src/main/paradox/query.md
+++ b/docs/src/main/paradox/query.md
@@ -71,8 +71,9 @@ query from a given `offset` after a crash/restart.
 
 The offset is a `TimestampOffset` and it is based on the database `transaction_timestamp()` when the event was stored.
 `transaction_timestamp()` is the time when the transaction started, not when it was committed. This means that a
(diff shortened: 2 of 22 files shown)


johanandren (approved):
LGTM!

docs/src/main/paradox/query.md
```diff
@@ -0,0 +1,136 @@
...
+query from a given `offset` after a crash/restart.
+
+The offset is a `TimestampOffset` and it is based on the database `transaction_timestamp()` when the event was stored.
+`transaction_timestamp()` is the time when the transaction started, not when it was committed. This means that a
+"later" event may be visible first and when retrieving events after the previously seen timestamp may miss some
+events. In distributed SQL databases there can also be clock skews for the database timestamps. For that reason
+`eventsBySlices` will perform additional backtracking queries to catch missed events. Events from backtracking
+will typically be duplicates of previously emitted events. It's the responsibility of the consumer to filter
+duplicates and make sure that events are processed in exact sequence number order for each persistence id.
+Such deduplication is provided by the R2DBC Projection.
```
octonato, line 79: What is maybe missing here is the ordering guarantees. 

The backtracking deliver missing events, but will we have events out of ordering? 
Yes, they can be out of ordering for the slice, but not for the same persistenceId. 

Maybe be good to clarify this.

docs/src/main/paradox/query.md
```diff
@@ -0,0 +1,136 @@
...
+
+Java
+:  @@snip [create](/docs/src/test/java/jdocs/home/query/QueryDocCompileOnly.java) { #currentChangesBySlices }
+
+Scala
+:  @@snip [create](/docs/src/test/scala/docs/home/query/QueryDocCompileOnly.scala) { #currentChangesBySlices }
+
+The emitted `DurableStateChange` can be a `UpdatedDurableState` or `DeletedDurableState`, but `DeletedDurableState` is not implemented yet.
+
+It will emit an `UpdatedDurableState` that when the durable state is updated. When the state is updated again another
```
octonato, line 121:
```suggestion
It will emit an `UpdatedDurableState` when the durable state is updated. When the state is updated again another
```

octonato (approved):
LGMT, with some suggestions

docs/src/main/paradox/durable-state-store.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Configuration
+
+To enable the journal plugin to be used by default, add the following line to your Akka `application.conf`:
+
+```
+akka.persistence.state.plugin = "akka.persistence.r2dbc.durable-state-store"
+```
+
+It can also be enabled with the `durableStateStorePluginId` for a specific `DurableStateBehavior` and multiple
```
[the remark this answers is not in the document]
patriknw: I think that belongs to the reference or api documentation of Akka. I don't think that is used often, so we shouldn't use too many words about that here.

docs/src/main/paradox/durable-state-store.md
```diff
@@ -0,0 +1,33 @@
+# Durable state store plugin
+
+The durable state plugin enables storing and loading key-value entries for @extref:[durable state actors](akka:typed/durable-state/persistence.html).
+
+## Schema
+
+The `durable_state` table and `durable_state_slice_idx` index need to be created in the configured database, see schema definition in @ref:[Creating the schema](getting-started.md#schema).
+
+The `durable_state_slice_idx` is only needed if the slice based @ref:[queries](query.md) are used.
```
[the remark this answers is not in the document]
patriknw: done

docs/src/main/paradox/query.md
```diff
@@ -0,0 +1,136 @@
...
+
+`eventsBySlices` should be used via Akka Projection, which will automatically handle the following
+difficulties.
+
+The consumer can keep track of its current position in the event stream by storing the `offset` and restart the
+query from a given `offset` after a crash/restart.
+
+The offset is a `TimestampOffset` and it is based on the database `transaction_timestamp()` when the event was stored.
+`transaction_timestamp()` is the time when the transaction started, not when it was committed. This means that a
+"later" event may be visible first and when retrieving events after the previously seen timestamp may miss some
```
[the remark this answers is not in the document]
patriknw: done

docs/src/main/paradox/query.md
patriknw: they can be out of order for a persistenceId because of the "missed events". I added more clarification of that in https://github.com/akka/akka-persistence-r2dbc/pull/149/commits/1fd44dfe7ecaccbb6cec7242c5053ad56b63d241

For example:
emitting 3, 4, 5
missing 6, 7
emitting 8, 9
backtracking emitting 5, 6, 7, 8, 9

meaning that a consumer will receive 8 before 6 and 7

docs/src/main/paradox/query.md
patriknw: done

docs/src/main/paradox/durable-state-store.md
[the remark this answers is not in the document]
octonato: Yeah, agree

docs/src/main/paradox/query.md
octonato: Hmm, I didn't know that it was possible to have out of order for the same persistence id. 
But isn't that so that the single write would make it unlikely, no?

patriknw: Out of order is a consequence of missed events, and missed events can happen when we move the offset forward (for another slice or pid). We can miss events when we switch to next "page" because we query from the latest offset.

PR #150, opened 2021-11-16
Tuning parameters

patriknw:
* adding two tuning parameters because I would like to be easily
  test the performance impact of these
* might be removed again, depending on result

### DIFF
--- a/core/src/test/scala/akka/persistence/r2dbc/state/DurableStateStoreSpec.scala
+++ b/core/src/test/scala/akka/persistence/r2dbc/state/DurableStateStoreSpec.scala
@@ -73,6 +73,9 @@ class DurableStateStoreSpec
     }
 
     "detect and reject concurrent updates" in {
+      if (!r2dbcSettings.durableStateAssertSingleWriter)
+        pending
+
       val entityType = nextEntityType()
       val persistenceId = PersistenceId(entityType, "id-to-be-updated-concurrently")
       val value = "Genuinely Collaborative"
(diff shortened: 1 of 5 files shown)


johanandren (approved):

core/src/test/scala/akka/persistence/r2dbc/state/DurableStateStoreSpec.scala
```diff
@@ -73,6 +73,9 @@ class DurableStateStoreSpec
     }
 
     "detect and reject concurrent updates" in {
+      if (!r2dbcSettings.durableStateAssertSingleWriter)
+        pending
```
octonato, line 77: why? 
Are we assessing the performance impact of it or are we chasing a bug? 

I'm fine with that. Sure there is a reason. I just want to understand.

octonato (approved):
LGTM

core/src/test/scala/akka/persistence/r2dbc/state/DurableStateStoreSpec.scala
octonato: To be more precise about my 'chasing bug', seems that we know about a flaky test that get solved when enabling single writer assertion. And that we want to look at that later and not let it break the builds for now.

patriknw: the config property for durableStateAssertSingleWriter in reference.conf is by default `true` so this test will run as before, but if we are running tests with that config set to `false` it would fail because it is testing exactly what is disabled

in other words, it's not pending because of any flakiness

I might remove these tuning parameters again but for now I want an easy way to toggle these when running in a real environment.

PR #159, opened 2021-11-23
Error when event from backtracking is rejected

patriknw:
* to get insight into if we need further improvement


projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -91,7 +95,11 @@ private[projection] object R2dbcProjectionImpl {
               s"Expected sourceProvider [${sourceProvider.getClass.getName}] " +
               "to implement LoadEventQuery when used with eventsBySlices.")
         }).map { loadedEnv =>
-          log.debug("Loaded event lazily, persistenceId [{}], seqNr [{}]", pid, seqNr)
+          val count = loadEnvelopeCounter.incrementAndGet()
+          if (count % 1000 == 0)
+            log.info("Loaded event lazily, persistenceId [{}], seqNr [{}]. Load count [{}]", pid, seqNr, count)
```
patriknw, line 100: this is to get insight to if many are lazy loaded

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -120,6 +140,29 @@ private[projection] object R2dbcProjectionImpl {
...
+   * otherwise they it would be stuck at the rejected event. Failing in that case so that we can notice if this is a
+   * problem. See issue #75.
+   */
+  private def checkBacktrackingNotAccepted[Envelope](env: Envelope): Unit = {
+    env match {
+      case eventEnvelope: EventEnvelope[_] if eventEnvelope.eventOption.isEmpty =>
+        throw new IllegalStateException(
+          s"Unexpected rejected event from backtracking, " +
+          s"persistenceId [${eventEnvelope.persistenceId}], seqNr [${eventEnvelope.sequenceNr}]. " +
+          "Please report this issue at https://github.com/akka/akka-persistence-r2dbc")
```
patriknw, line 154: this will result in projection restart, and logging

patriknw:
I'll rework this a bit

PR #163, opened 2021-11-24
config setting for logging of (slow) db calls

patriknw:
it can also be good to know that logging of slow queries can be enabled on the Yugabyte server side: https://docs.yugabyte.com/latest/explore/query-1-performance/query-tuning-intro/#log-slow-queries


johanandren (approved):

octonato (approved):
LGTM

Issue #106, opened 2021-11-02
Failed to delete timestamp offset


patriknw:
Noticed this when running akka-projection-testing for a longer time with Yugabyte:
```
[2021-11-02 11:53:14,157] [WARN] [akka.projection.r2dbc.internal.R2dbcOffsetStore] [] [] [test-akka.actor.default-dispatcher-27] [] - Failed to delete timestamp offset until [2021-11-02T11:46:34.550356Z] for projection [test-projection-id-0-projection-0-tag-2]: io.r2dbc.postgresql.ExceptionFactory$PostgresqlRollbackException: [40001] Query error: Restart read required at: { read: { physical: 1635853993768562 } local_limit: { physical: 1635853994138972 } global_limit: <min> in_txn_limit: <max> serial_no: 0 }  MDC: {}
```

patriknw:
The delete is also rather slow, and I guess that could be a reason for why it often has to be "restarted".

```
explain analyze delete from akka_projection_timestamp_offset_store WHERE slice BETWEEN 0 and 31 AND projection_name = 'test-projection-id-0' AND timestamp_offset < '2021-11-22 16:58:21.801179+00';
                                                                                            QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Delete on akka_projection_timestamp_offset_store  (cost=0.00..16.25 rows=100 width=32) (actual time=349.157..349.157 rows=0 loops=1)
   ->  Index Scan using akka_projection_timestamp_offset_store_pkey on akka_projection_timestamp_offset_store  (cost=0.00..16.25 rows=100 width=32) (actual time=349.155..349.155 rows=0 loops=1)
         Index Cond: ((slice >= 0) AND (slice <= 31) AND ((projection_name)::text = 'test-projection-id-0'::text))
         Filter: (timestamp_offset < '2021-11-22 16:58:21.801179+00'::timestamp with time zone)
         Rows Removed by Filter: 1
 Planning Time: 0.098 ms
 Execution Time: 349.219 ms
 ```
 
 We could use an append only approach instead of upserts. Then the primary key could be like this:
 ```
 CREATE TABLE IF NOT EXISTS akka_projection_timestamp_offset_store2 (
  projection_name VARCHAR(255) NOT NULL,
  projection_key VARCHAR(255) NOT NULL,
  slice INT NOT NULL,
  persistence_id VARCHAR(255) NOT NULL,
[...]

PR #164, opened 2021-11-25
Insert offset instead of upsert, #106

patriknw:
* should make the deletes <= timestamp faster since
  it can use the prefix of the primary key
* note that primary key of akka_projection_timestamp_offset_store
  was changed, so existing tables must be recreated (can be populated
  with old data with 'insert into .. select * from'

References #106

### DIFF
--- a/ddl-scripts/create_tables_postgres.sql
+++ b/ddl-scripts/create_tables_postgres.sql
@@ -74,9 +74,9 @@ CREATE TABLE IF NOT EXISTS akka_projection_timestamp_offset_store (
   seq_nr BIGINT NOT NULL,
   -- timestamp_offset is the db_timestamp of the original event
   timestamp_offset timestamp with time zone NOT NULL,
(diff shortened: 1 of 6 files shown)


ddl-scripts/create_tables_postgres.sql
```diff
@@ -77,7 +77,7 @@ CREATE TABLE IF NOT EXISTS akka_projection_timestamp_offset_store (
   -- last_updated is when the offset was stored
   -- the consumer lag is last_updated - timestamp_offset
   last_updated timestamp with time zone NOT NULL,
```
johanandren, line 79: This column name is a bit weird now

johanandren (approved):

ddl-scripts/create_tables_postgres.sql
patriknw: ah, I'll change

johanandren (approved):

Issue #77, opened 2021-10-26
Combine with live stream of events sent as Akka messages


patriknw:
With the precise deduplication in place I think we have good opportunity for publishing the events over Distributed PubSub to the projection streams.

It should only be fire-n-forget publishing and not any attempts of reliable delivery.

Then we can reduce the database polling frequency.

patriknw:
Done in https://github.com/akka/akka-persistence-r2dbc/pull/168

PR #168, opened 2021-11-30
Send events over pub-sub, #77

patriknw:
* When lower latency for consuming in Projection is needed the envelopes can
  be sent over pub-sub to the eventsBySlices queries.
* De-duplication by the offset store in the same way as before.

Missing serializer for EventEnvelope in Akka.

This is a hackday experiment, but I don't think much more than some reference docs is needed.

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/Sql.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/Sql.scala
@@ -9,14 +9,14 @@ import scala.annotation.varargs
 import akka.annotation.InternalStableApi
 
 /**
- * INTERNAL API: Utility to format SQL strings. Replaces `?` with numbered `$1`, `$2` for bind parameters. Trims
+ * INTERNAL API: Utility to format SQL strings. Replaces `?` with numbered `\$1`, `\$2` for bind parameters. Trims
  * whitespace, including line breaks.
  */
 @InternalStableApi
 object Sql {
 
   /**
-   * Scala string interpolation with `sql` prefix. Replaces `?` with numbered `$1`, `$2` for bind parameters. Trims
+   * Scala string interpolation with `sql` prefix. Replaces `?` with numbered `\$1`, `\$2` for bind parameters. Trims
    * whitespace, including line breaks. Standard string interpolation arguments `$` can be used.
    */
   implicit class Interpolation(val sc: StringContext) extends AnyVal {
@@ -25,7 +25,7 @@ object Sql {
   }
 
   /**
-   * Java API: Replaces `?` with numbered `$1`, `$2` for bind parameters. Trims whitespace, including line breaks. The
+   * Java API: Replaces `?` with numbered `\$1`, `\$2` for bind parameters. Trims whitespace, including line breaks. The
    * arguments are used like in [[java.lang.String.format]].
    */
   @varargs


projection/src/test/scala/akka/projection/r2dbc/EventSourcedPubSubSpec.scala
```diff
@@ -0,0 +1,240 @@
...
+              .asInstanceOf[TimestampOffset]
+              .readTimestamp)
+        log.info("via pub-sub {}: {}", pid, viaPubSub.map(_.envelope.sequenceNr).mkString(", "))
+      }
+
+      val countViaPubSub = processed.count(p =>
+        p.envelope.offset.asInstanceOf[TimestampOffset].timestamp == p.envelope.offset
+          .asInstanceOf[TimestampOffset]
+          .readTimestamp)
+      log.info("Total via pub-sub: {}", countViaPubSub)
```
patriknw, line 233: The log output looks like this. Note that the test is simulating dropped messages by using a small buffer size.

```
via pub-sub TestEntity-1|p7: 2, 3
via pub-sub TestEntity-1|p17: 1, 2, 10
via pub-sub TestEntity-1|p6: 2, 3, 4
via pub-sub TestEntity-1|p10: 2, 3
via pub-sub TestEntity-1|p13: 1, 2, 10
via pub-sub TestEntity-1|p4: 2, 3, 4
via pub-sub TestEntity-1|p9: 2, 3, 4, 5, 6, 7
via pub-sub TestEntity-1|p18: 1, 2, 3, 10
via pub-sub TestEntity-1|p0: 1, 2, 3, 4, 5, 6, 7, 8, 10
via pub-sub TestEntity-1|p5: 2, 3, 4, 5
via pub-sub TestEntity-1|p15: 1, 2, 10
via pub-sub TestEntity-1|p1: 2, 3, 4
via pub-sub TestEntity-1|p2: 2, 3, 4
via pub-sub TestEntity-1|p11: 1, 2, 3, 10
via pub-sub TestEntity-1|p16: 1, 2, 3, 10
via pub-sub TestEntity-1|p14: 1, 2, 10
via pub-sub TestEntity-1|p3: 2, 3, 4
via pub-sub TestEntity-1|p19: 1, 2, 10
via pub-sub TestEntity-1|p12: 1, 2, 10
via pub-sub TestEntity-1|p8: 2, 3
Total via pub-sub: 70
```

With default (large) buffer size:

```
via pub-sub TestEntity-1|p7: 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p17: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p6: 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p10: 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p13: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p4: 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p9: 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p18: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p0: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p5: 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p15: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p1: 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p2: 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p11: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p16: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p14: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p3: 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p19: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p12: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
via pub-sub TestEntity-1|p8: 2, 3, 4, 5, 6, 7, 8, 9, 10
Total via pub-sub: 190
```

core/src/main/scala/akka/persistence/r2dbc/internal/PubSub.scala
```diff
@@ -0,0 +1,69 @@
...
+  // Java API
+  def get(system: ActorSystem[_]): PubSub = apply(system)
+
+}
+
+/**
+ * INTERNAL API
+ */
+@InternalApi private[akka] class PubSub(system: ActorSystem[_]) extends Extension {
+  private val topics = new ConcurrentHashMap[String, ActorRef[Any]]
```
johanandren, line 39: Could it not be `ActorRef[Topic.Command[EventEnvelope[Event]]]` here already and skip the unsafeUpcast.narrow below?

johanandren (commented):
Neat!

core/src/main/scala/akka/persistence/r2dbc/internal/PubSub.scala
patriknw: yes, it can have some more type, but Event is unknown

johanandren: Ofc, I missed that it is a method type param.

core/src/main/scala/akka/persistence/r2dbc/journal/R2dbcJournal.scala
```diff
@@ -164,8 +171,26 @@ private[r2dbc] final class R2dbcJournal(config: Config, cfgPath: String) extends
         atomicWrite(batch).map(_ => Nil)(ExecutionContexts.parasitic)
       }
     writesInProgress.put(persistenceId, writeResult)
-    writeResult.onComplete { _ =>
-      context.self ! WriteFinished(persistenceId, writeResult)
+    writeResult.onComplete {
+      case Success(_) if pubSub.isDefined =>
+        val seqNr = messages.head.lowestSequenceNr
+        query
+          .timestampOf(persistenceId, seqNr)
```
octonato, line 178: I believe we can remove this extra query if we use inserts with `returning` syntax. 

For example:

```sql
postgres=# create table foo (
name VARCHAR(255) not null
);
CREATE TABLE

postgres=# INSERT INTO foo(name) VALUES ('bar') returning transaction_timestamp();
     transaction_timestamp
-------------------------------
 2021-12-02 19:58:00.491517+00
(1 row)
```

This syntax is only for postgres, so for other databases (if we ever add support for others) we could then run that extra query. 

For batch inserts, it will return a list of intervals, but we only need one anyway since they will all be the same.

octonato (commented):
Really nice! 

Left a comment about a possible optimization for inserts.

core/src/main/scala/akka/persistence/r2dbc/journal/R2dbcJournal.scala
patriknw: Thanks for the suggestion, I have implemented that in https://github.com/akka/akka-persistence-r2dbc/pull/168/commits/a6d4ada76bfb67361ff78f811004a3c0afe93099

patriknw:
@johanandren @octonato I have completed this, including docs

core/src/main/scala/akka/persistence/r2dbc/internal/PubSub.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  // Java API
+  def get(system: ActorSystem[_]): PubSub = apply(system)
+
+}
+
+/**
+ * INTERNAL API
+ */
+@InternalApi private[akka] class PubSub(system: ActorSystem[_]) extends Extension {
```
johanandren, line 38: Looking at this again I wonder if we should perhaps have a small convenience TopicRegistry extension in Akka to make it easy to run one topic actor per topic name in the local actor system rather than having to keep track of those yourself.

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -102,9 +103,9 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
 
     val insertEventWithTransactionTimestampSql = {
       if (journalSettings.dbTimestampMonotonicIncreasing)
-        sql"$baseSql transaction_timestamp())"
+        sql"$baseSql transaction_timestamp()) RETURNING transaction_timestamp() AS db_timestamp"
       else
-        sql"$baseSql GREATEST(transaction_timestamp(), $timestampSubSelect))"
+        sql"$baseSql GREATEST(transaction_timestamp(), $timestampSubSelect)) RETURNING transaction_timestamp() AS db_timestamp"
```
johanandren, line 108: Doesn't this mean it could insert the subselect one but return the transaction timestamp -> mismatch?

core/src/main/scala/akka/persistence/r2dbc/journal/R2dbcJournal.scala
```diff
@@ -147,27 +157,62 @@ private[r2dbc] final class R2dbcJournal(config: Config, cfgPath: String) extends
...
       else {
         // persistAsync case
         // easiest to just group all into a single AtomicWrite
         val batch = AtomicWrite(messages.flatMap(_.payload))
-        atomicWrite(batch).map(_ => Nil)(ExecutionContexts.parasitic)
+        atomicWrite(batch)
+      }
+
+    val writeAndPublishResult: Future[Done] =
+      publish(messages, writeResult)
```
johanandren, line 178: Doesn't this mean we could publish even if write fails? Shouldn't it be onComplete for writeResult rather?

patriknw: inside `def publish` it is `flatMap` on the `writeResult`

patriknw: I can rename that `publish` to `publishAfterWrite`

johanandren: Ah, the flatMap inside `publish` is what I was missing. I think it's ok like this.

core/src/main/scala/akka/persistence/r2dbc/internal/PubSub.scala
johanandren: Created https://github.com/akka/akka/issues/31053 to track that idea on the Akka side.

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
octonato: `RETURNING db_timestamp` should work

octonato (commented):
I will look in details tomorrow, but for the sub-select, returning the column should, if I remember correctly.

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -180,31 +183,54 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
 
-    if (log.isDebugEnabled())
-      result.foreach { updatedRows =>
-        log.debug("Wrote [{}] events for persistenceId [{}]", updatedRows, events.head.persistenceId)
+        if (log.isDebugEnabled())
+          result.foreach { updatedRows =>
+            log.debug("Wrote [{}] events for persistenceId [{}]", updatedRows, events.head.persistenceId)
+          }
+
+        result.map(_ => None)(ExecutionContexts.parasitic)
```
octonato, line 210: Returning should also work with batch inserts. Have you tried it? 

The annoying part is that it will return a list  of timestamps as large as the number of items in the batch. 

For example, 

```sql
postgres=# INSERT INTO foo(name) VALUES ('abc'), ('def') returning transaction_timestamp();

     transaction_timestamp
-------------------------------
 2022-01-12 13:10:41.451654+00
 2022-01-12 13:10:41.451654+00
(2 rows)

INSERT 0 2
```

octonato (approved):
LGTM, but I still have questions / thoughts about batching and `returning`.

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
patriknw: thanks, changed to that

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -180,31 +183,54 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
+          }
       }
 
-    result.map(_ => ())(ExecutionContext.parasitic)
+      if (log.isDebugEnabled())
+        result.foreach { updatedRows =>
+          log.debug("Wrote [{}] events for persistenceId [{}]", updatedRows, events.head.persistenceId)
+        }
+
+      result.map(_ => Some(events.head.dbTimestamp))(ExecutionContext.parasitic)
```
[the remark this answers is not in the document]
patriknw: simplified these cases

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
patriknw: yup, changed to returning for the batch case also

octonato (approved):
LGTM

johanandren (approved):

patriknw:
For the record, I released `v0.4.1` before merging this if we want to compare performance of the new statements

PR #170, opened 2021-12-04
Utility to replace numbered sql parameters, #135

### LINKED ISSUE #135: Utility to replace numbered sql parameters

patriknw:
Pretty neat.

* Replaces `?` with numbered `$1`, `$2` for bind parameters.
* Trims whitespace, including line breaks.

Fixes #135

### DIFF
--- a/docs/src/main/paradox/projection.md
+++ b/docs/src/main/paradox/projection.md
@@ -169,16 +169,6 @@ Java
 Such simple handlers can also be defined as plain functions via the helper @scala[`R2dbcHandler.apply`]@java[`R2dbcHandler.fromFunction`] factory method.
 @@@
 
-@@@ note
-
-R2DBC bind parameters must be defined as `$1`, `$2`, `$3`. The numbering can become inconvenient and doesn't
-improve readability.
-@scala[The string interpolation utility in @apidoc[Sql.Interpolation](akka.persistence.r2dbc.Sql.Interpolation)]
-@java[The utility @apidoc[Sql.format](akka.persistence.r2dbc.Sql$)] is useful for this purpose.
-It Replaces `?` with numbered `$1`, `$2` for bind parameters. Trims whitespace, including line breaks.
-
-@@@
-
 ### Grouped handler
 
 When using @ref:[`R2dbcProjection.groupedWithin`](#groupedwithin) the handler is processing a @scala[`Seq`]@java[`List`] of envelopes.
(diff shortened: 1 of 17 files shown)


docs/src/main/paradox/projection.md
```diff
@@ -169,6 +169,16 @@ Java
 Such simple handlers can also be defined as plain functions via the helper @scala[`R2dbcHandler.apply`]@java[`R2dbcHandler.fromFunction`] factory method.
 @@@
 
+@@@ note
+
+R2DBC bind parameters must be defined as `$1`, `$2`, `$3`. The numbering can become inconvenient and doesn't
+improve readability.
+@scala[The string interpolation utility in @apidoc[Sql.Interpolation](akka.persistence.r2dbc.Sql.Interpolation)]
+@java[The utility @apidoc[Sql.format](akka.persistence.r2dbc.Sql$)] is useful for this purpose.
+It Replaces `?` with numbered `$1`, `$2` for bind parameters. Trims whitespace, including line breaks.
```
octonato, line 178: I don't think this limitation is imposed by R2DBC. On their docs they let know that it's vendor-specific and that they don't have the goal to address it (see https://r2dbc.io/spec/0.9.0.RELEASE/spec/html/#statements.parameterized). 

So, basically, the `$num` syntax is Postgres specific. 

It's great to have an interpolation utility though. But since we are introducing it, we may want to take one little step further and have named parameters instead. That's the most annoying part right now because we need to keep track of the indices and with all the dynamic queries we have, it's quite annoying.

So, I was thinking on something like: 

val stmt = sql"""INSERT INTO foo (name, address) VALUES ($name, address)"""

Followed by named binding calls: 

```scala
stmt
  .bind("address", "Sesame Street")
  .bind("name", "Ernie")
```

The ordering of the binding doesn't matter. Behind the scenes we translate then to the real binding call with the right index. 

Then, per dialect, we will need to replace the names by whatever notation the dialect requires, but the logic to order the bindings back to the real indices will be the same for all dialects.

octonato (approved):
Neat indeed. 

Interpolation is such a cool thing. We should do more of it. :-)

docs/src/main/paradox/projection.md
patriknw: so stmt would be our own class, with a HashMap lookup?

octonato: Yes, something like that. 

Some indirection, but it will bring it to the next level. :-)

patriknw: ok, I will not make that addition of named parameters here. I'll move it to internal so that we have more freedom to change.

johanandren (approved):

Issue #90, opened 2021-10-26
DAO layer for storing query representation of DurableState


patriknw:
Can we right away implement a DAO layer (also with Java support) for customizing a table structure for storing typed data and adapting queries for this structure?

_Originally posted by @ihostage in https://github.com/akka/akka-persistence-r2dbc/issues/78#issuecomment-951909109_

PR #171, opened 2021-12-05
API sketch of DAO extension for durable state, #90

patriknw:
* for storing query representation from the write side
* in same transaction as the upsert
* support for additional columns in the durable_state table, for secondary indexes
* support for any additional statements for storing more complex
  representations, such as in a secondary table

@ihostage would this cover what you requested in https://github.com/akka/akka-persistence-r2dbc/issues/90 ?

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -79,6 +79,13 @@ akka.persistence.r2dbc {
     # previous revision. There might be a small performance gain if
     # this is disabled.
     assert-single-writer = on
+
+    # Fully qualified class name of a DaoExtension that can define
+    # additional processing for Durable State upserts or deletes.
+    # The class must implement akka.persistence.r2dbc.state.javadsl.DaoExtension
+    # or akka.persistence.r2dbc.state.scaladsl.DaoExtension.
+    # It may optionally define an ActorSystem constructor parameter.
+    dao-extension = ""
   }
 }
 // #durable-state-settings
--- a/core/src/main/scala/akka/persistence/r2dbc/state/javadsl/DaoExtension.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/state/javadsl/DaoExtension.scala
@@ -1,0 +1,74 @@
+/*
+ * Copyright (C) 2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.r2dbc.state.javadsl
+
+import java.util.Collections
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.CompletionStage
+import java.util.{ List => JList }
+import java.util.function.{ Function => JFunction }
+
+import akka.Done
+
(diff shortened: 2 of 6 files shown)


ihostage:
Hi, Patrik! 👋 
> @ihostage would this cover what you requested in #90 ?

Yes, but partially 😄 
On this extension, I can't completely get rid of the need to store a blob representation of the entity. But it gives me the opportunity to store typed information and index them for effectually use on the "read-side". Thanks a lot 👍

patriknw:
> On this extension, I can't completely get rid of the need to store a blob representation of the entity. But it gives me the opportunity to store typed information and index them for effectually use on the "read-side".

My thinking is that completely re-implementing the DAO is too much for most users. That would not only involve implementing the storage, but also queries. With this approach of "just" adding secondary index columns the existing functionality will continue working.

We might add an even more powerful DAO extension point for implementing a different dialect for a different database backend. If you don't want the blob storage it sounds more like a new dialect.

Alternatively, you can store the other representation in a separate table via the `processOnUpsert`.

octonato:
> My thinking is that completely re-implementing the DAO is too much for most users. 

I can understand that, but the users will need to get their hands dirty anyway because they will need to build DAOs to read data using their custom indices. So, at the end of the day, they will have yet another DAO and will need to get to know how to do work with R2DBC. 

Another problem with this approach is that it only works if I have only one type of durable actor. 

Btw, I think the complexity of implementing a custom DAO is not on the implementation of `insert`, `update` and `delete`, but on the queries by slice (and friends). 

If I chose to have custom DAO, I'm probably looking for a CRUD style persistence and therefore I don't care about `changesBySlices` or any other built-in queries. I will be querying using my custom indices and nothing else.

patriknw:
> Another problem with this approach is that it only works if I have only one type of durable actor.

That's why I defined `def tableName(entityType: String)` so that tables with different columns can be used for different types of durable actor.

octonato:
> That's why I defined `def tableName(entityType: String)` so that tables with different columns can be used for different types of durable actor.

Oh, I missed this. Sorry.

patriknw:
> I can understand that, but the users will need to get their hands dirty anyway because they will need to build DAOs to read data using their custom indices. So, at the end of the day, they will have yet another DAO and will need to get to know how to do work with R2DBC.

but that is no different from ordinary Projections. The `def processOnUpsert` is pretty much the same as a Projection handler and can store the projected representation in whatever way they choose, but still not break other existing functionality of the plugin.

> If I chose to have custom DAO, I'm probably looking for a CRUD style persistence and therefore I don't care about changesBySlices or any other built-in queries. I will be querying using my custom indices and nothing else.

I don't fully agree. Other queries may still be relevant, for example to also use ordinary Projections, e.g. publish events to Kafka. Another mandatory query is eventsByPersistenceId, since that is used by ordinary replay (and for Replicated Event Sourcing).

PR #177, opened 2021-12-14
Remove slice from primary key, #175

### LINKED ISSUE #175: Slice hashing thoughts
For the journal table we have a HASH partition key:
```
PRIMARY KEY((slice, entity_type) HASH, persistence_id, seq_nr ASC)
```

One potential problem is that we have rather few slices so there is a risk that the rows are not evenly distributed over the tablets (db sharding).

Given the table:
```
CREATE TABLE IF NOT EXISTS test_journal(
  slice INT NOT NULL,
  entity_type VARCHAR(255) NOT NULL,
  persistence_id VARCHAR(255) NOT NULL,
  seq_nr BIGINT NOT NULL,
  db_timestamp timestamp with time zone NOT NULL,
  event VARCHAR(255),

  PRIMARY KEY((slice, entity_type) HASH, persistence_id, seq_nr ASC)
);
```

and inserting 1 row per slice:
```
DO
$do$
BEGIN
  for i in 0..127 loop
    insert into test_journal (slice, entity_type, persistence_id, seq_nr, db_timestamp, event)
    values (i, 'MyEntity', 'pid-' || i, 1, transaction_timestamp(), 'event-' || i);
  end loop;
END
$do$;
```

Yugabyte is using 2 bytes for the shard hash code, i.e. the range is 0-65535. By default it slits a table in 8 tablets per node, i.e. 24 tablets per table so let's experiment with that value. That means that each tablet takes a range of 2730 hash values.

We can look at the distributions with the following sql:
```
select slice, yb_hash_code(slice::int, entity_type::VARCHAR(255)) as h, yb_hash_code(slice::int, entity_type::VARCHAR(255)) / 2730 as t from test_journal order by h;

select yb_hash_code(slice::int, entity_type::VARCHAR(255)) / 2730 as t, count(*) from test_journal group by t order by t;

create table slice_distribution as (select yb_hash_code(slice::int, entity_type::VARCHAR(255)) / 2730 as t, count(*) from test_journal group by t order by t);

select min(count), max(count), min(count)::decimal * 100 / sum(count) as min_percent, max(count)::decimal * 100 / sum(count) as max_percent from slice_distribution;

select * from slice_distribution order by count;
 t  | count
----+-------
  4 |     2
  6 |     2
 17 |     2
 23 |     3
  3 |     3
 15 |     4
  8 |     4
 13 |     4
 16 |     4
 19 |     4
 14 |     5
 12 |     5
 18 |     6
  0 |     6
 11 |     6
 22 |     6
 10 |     6
  9 |     7
  5 |     7
  7 |     8
  1 |     8
  2 |     8
 20 |     9
 21 |     9
```

That doesn't look great. Some get 2 of 128 slices, and some 9 of 128.

One can repeat the experiment with 1024 slices, and it becomes better but would that be good enough?

```
t  | count
----+-------
 23 |    34
  8 |    34
  4 |    35
 12 |    37
  6 |    38
 22 |    38
 14 |    39
 17 |    39
 10 |    40
 21 |    42
 15 |    42
 13 |    42
  9 |    43
 19 |    43
  1 |    44
  0 |    44
 11 |    45
  3 |    45
  2 |    49
  7 |    49
 16 |    49
  5 |    49
 20 |    51
 18 |    53
 ```
 
 ## Why slice as sharding key for journal table?
 
 We don't have to use the slice as sharding key in the journal table. We could use the `persistence_id` instead and that would give good distribution assuming that persistence_id is random.
 
 The slice is needed for the `eventsBySlices` query (Projections) but for that we have the slice index.
 
I started out with the slice as sharding key in the journal table because I thought it was a "natural" sharding key that we anyway had in place. I also found the "Co-partitioning" feature in https://github.com/yugabyte/yugabyte-db/blob/master/architecture/design/ysql-tablegroups.md which could make the index and the parent table colocated if they have the same sharding key. That feature doesn't exist yet.

`eventsBySlices` would search the index and then retrieve the event payload for matching entries from the parent journal table. My thinking was that such fetch would be most efficient if it would be in a batch request from one (a few) tablets rather than spread over all tablets. That's just speculation, though.

## Why use HASH sharding for journal table?

Most recommendations in the Yugabyte documentation prefer hash sharding but maybe range sharding would be equally good for the slice in the journal table.

### PR DESCRIPTION
**Pull request by patriknw**
* see discussion and testing in issue #175

Fixes #175

cc @jrudolph

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
@@ -77,7 +77,7 @@ private[r2dbc] final class SnapshotDao(settings: R2dbcSettings, connectionFactor
     INSERT INTO $snapshotTable
     (slice, entity_type, persistence_id, seq_nr, write_timestamp, snapshot, ser_id, ser_manifest, meta_payload, meta_ser_id, meta_ser_manifest)
     VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
-    ON CONFLICT (slice, entity_type, persistence_id)
+    ON CONFLICT (persistence_id)
     DO UPDATE SET
       seq_nr = excluded.seq_nr,
       write_timestamp = excluded.write_timestamp,
@@ -108,7 +108,7 @@ private[r2dbc] final class SnapshotDao(settings: R2dbcSettings, connectionFactor
     sql"""
       SELECT persistence_id, seq_nr, write_timestamp, snapshot, ser_id, ser_manifest, meta_payload, meta_ser_id, meta_ser_manifest
       FROM $snapshotTable
-      WHERE slice = ? AND entity_type = ? AND persistence_id = ?
+      WHERE persistence_id = ?
       $maxSeqNrCondition $minSeqNrCondition $maxTimestampCondition $minTimestampCondition
       LIMIT 1"""
   }
@@ -132,24 +132,19 @@ private[r2dbc] final class SnapshotDao(settings: R2dbcSettings, connectionFactor
 
     sql"""
       DELETE FROM $snapshotTable
-      WHERE slice = ? AND entity_type = ? AND persistence_id = ?
+      WHERE persistence_id = ?
       $maxSeqNrCondition $minSeqNrCondition $maxTimestampCondition $minTimestampCondition"""
   }
 
   def load(persistenceId: String, criteria: SnapshotSelectionCriteria): Future[Option[SerializedSnapshotRow]] = {
-    val entityType = PersistenceId.extractEntityType(persistenceId)
-    val slice = persistenceExt.sliceForPersistenceId(persistenceId)
-
     r2dbcExecutor
       .select(s"select snapshot [$persistenceId], criteria: [$criteria]")(
         { connection =>
           val statement = connection
             .createStatement(selectSql(criteria))
-            .bind(0, slice)
-            .bind(1, entityType)
-            .bind(2, persistenceId)
+            .bind(0, persistenceId)
 
-          var bindIdx = 2
+          var bindIdx = 0
           if (criteria.maxSequenceNr != Long.MaxValue) {
             bindIdx += 1
             statement.bind(bindIdx, criteria.maxSequenceNr)
@@ -217,11 +212,9 @@ private[r2dbc] final class SnapshotDao(settings: R2dbcSettings, connectionFactor
     r2dbcExecutor.updateOne(s"delete snapshot [$persistenceId], criteria [$criteria]") { connection =>
       val statement = connection
         .createStatement(deleteSql(criteria))
-        .bind(0, slice)
-        .bind(1, entityType)
-        .bind(2, persistenceId)
+        .bind(0, persistenceId)
 
-      var bindIdx = 2
+      var bindIdx = 0
       if (criteria.maxSequenceNr != Long.MaxValue) {
         bindIdx += 1
         statement.bind(bindIdx, criteria.maxSequenceNr)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: ddl-scripts/create_tables_yugabyte.sql [hunk 1]
```diff
@@ -17,12 +17,12 @@ CREATE TABLE IF NOT EXISTS event_journal(
...
   meta_payload BYTEA,
 
-  PRIMARY KEY((slice, entity_type) HASH, persistence_id, seq_nr ASC)
+  PRIMARY KEY(persistence_id HASH, seq_nr ASC)
 );
 
 -- `event_journal_slice_idx` is only needed if the slice based queries are used
-CREATE INDEX IF NOT EXISTS event_journal_slice_idx ON event_journal(entity_type ASC, slice ASC, db_timestamp ASC)
-  INCLUDE (seq_nr, deleted);
+CREATE INDEX IF NOT EXISTS event_journal_slice_idx ON event_journal(slice ASC, entity_type ASC, db_timestamp ASC)
```
patriknw, line 24: the order of the slice and entity_type columns shouldn't matter much, but there was already some inconsistency with the postrgres definition, and this makes it easier to pre-split into more tablets:
```
SPLIT AT VALUES ((31), (63), (95))
```

core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
```diff
@@ -132,24 +132,19 @@ private[r2dbc] final class SnapshotDao(settings: R2dbcSettings, connectionFactor
 
     sql"""
       DELETE FROM $snapshotTable
-      WHERE slice = ? AND entity_type = ? AND persistence_id = ?
+      WHERE persistence_id = ?
```
johanandren, line 135: I get removing the slice, but doesn't removing the `entity_type` make it so you can only ever store a single type of entity with the same persistence id, or do we expect that it will always be in the pid (what if someone is using the journal without sharding for example)? Worth a mention in the docs at least?

johanandren (approved):

core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
patriknw: The `persistenceId` must be unique across different entity types. That's historical reasons. Entity type was added much later. That's why we create the persistenceId string with the `PersistenceId.apply(entityTypeHint, entityId)` and it becomes `entityTypeHint|entityId`.

It's the same in all other persistence plugins.

The code here to retrieve the entityType was to extract it from the persistenceId.

ddl-scripts/create_tables_yugabyte.sql
octonato: Are we missing the `SPLIT` definiton? 

If I understand correctly, using split we instruct Yugabyte to distribute the data into the tablets according to the slice distribution we want. Is that correct? 

What will be the effect of it if we want to change the range we query (narrowing it)? I guess we will have more than one query hitting the same tablet. 
Would we get better performance if we re-create the index if we decide to change the range?

core/src/main/scala/akka/persistence/r2dbc/snapshot/SnapshotDao.scala
johanandren: I see, I didn't know, thanks for the clarification!

ddl-scripts/create_tables_yugabyte.sql
patriknw: Without the `SPLIT AT` definition it will use one tablet. That is probably fine as a default for an index table. At least we don't know for sure that it's a bottleneck yet so we should stick with that default.

Tablets can be split afterwards, with re-creating the index. There is also auto-splitting.

PR #179, opened 2021-12-15
Avoid repeatly trying to delete offsets when idle

patriknw:
* small resource usage optimization

### DIFF
--- a/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -693,44 +693,49 @@ private[projection] class R2dbcOffsetStore(
   }
 
   def deleteOldTimestampOffsets(): Future[Int] = {
-    val currentState = getState()
-    if (idle.getAndSet(true) || currentState.window.compareTo(settings.timeWindow) < 0) {
-      // no new offsets stored since previous delete, or it hasn't filled up the window yet
+    if (idle.getAndSet(true)) {
+      // no new offsets stored since previous delete
       Future.successful(0)
     } else {
-      val until = currentState.latestTimestamp.minus(settings.timeWindow)
-      val minSlice = timestampOffsetBySlicesSourceProvider.minSlice
-      val maxSlice = timestampOffsetBySlicesSourceProvider.maxSlice
-      val result = r2dbcExecutor.updateOne("delete timestamp offset") { conn =>
-        conn
-          .createStatement(deleteTimestampOffsetSql)
-          .bind(0, minSlice)
-          .bind(1, maxSlice)
-          .bind(2, projectionId.name)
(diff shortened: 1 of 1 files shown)


octonato (dismissed):
LGTM

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -688,8 +693,8 @@ private[projection] class R2dbcOffsetStore(
 
   def deleteOldTimestampOffsets(): Future[Int] = {
     val currentState = getState()
-    if (currentState.window.compareTo(settings.timeWindow) < 0) {
-      // it hasn't filled up the window yet
+    if (idle.getAndSet(true) || currentState.window.compareTo(settings.timeWindow) < 0) {
+      // no new offsets stored since previous delete, or it hasn't filled up the window yet
```
octonato, line 697: Hmm, I'm not sure if the logic here is correct. 

If idle is `false`, the first if branch won't pass, but we side-effect the atomic boolean. If the second branch pass the test, we won't be running any delete and now we put idle back to `true`.

octonato: Well, I guess in that case we lose one chance to run the deletion, but it will come back later anyway.

patriknw: thanks, I changed it: https://github.com/akka/akka-persistence-r2dbc/pull/179/commits/b934ee387a970adba0729dffe5f443e8185155da

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -687,43 +693,49 @@ private[projection] class R2dbcOffsetStore(
...
-          .bind(0, minSlice)
-          .bind(1, maxSlice)
-          .bind(2, projectionId.name)
-          .bind(3, until)
-      }
+      val currentState = getState()
+      if (currentState.window.compareTo(settings.timeWindow) < 0) {
+        // it hasn't filled up the window yet
+        Future.successful(0)
+      } else {
```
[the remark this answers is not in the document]
patriknw: Your reasoning is missing one thing here:

> if window not filled, return Future.successful(0), but at this point we side-effected the AtomicBoolean already, we may want to put it back to false?

The window will not change if there are no changes to the currentState. There must be some more offsets stored for the window to change and then we are back to idle=false.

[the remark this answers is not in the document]
octonato: I see, we don't need indeed to put it back to false as it will happen at some point.

octonato (approved):
LGTM

PR #180, opened 2021-12-16
Adjust log-db-calls-exceeding config

patriknw:


core/src/main/resources/reference.conf
```diff
@@ -137,7 +137,7 @@ akka.persistence.r2dbc {
   # Logs database calls that take longer than this duration at INFO level.
   # Set to "off" to disable this logging.
   # Set to 0 to log all calls.
-  log-db-calls-exceeding = 100 ms
+  log-db-calls-exceeding = 300 ms
```
patriknw, line 140: it was too verbose as default

johanandren (approved):

octonato (approved):
LGTM

PR #183, opened 2022-01-04
Split slice index, #174

### LINKED ISSUE #174: Revisit potential slice index bottleneck on yugabyte
Running performance tests on yugabyte we observed that the slice index seems to be a bottleneck, i.e. many or most requests at a time going to a single tablet.

This might be due to the fact, that we currently define the index to use ["range sharding"](https://docs.yugabyte.com/latest/architecture/docdb-sharding/sharding/) in this definition:

```sql
CREATE INDEX IF NOT EXISTS event_journal_slice_idx ON event_journal(entity_type ASC, slice ASC, db_timestamp ASC)
  INCLUDE (seq_nr, deleted);
```

Since `entity_type` and `slice` are usually of low cardinality it should be possible to change these at least to use `hash sharding`, instead, which might distribute the load better.

We should revisit what we require from the index both in terms of read and write performance and structure the keys / sharding accordingly.

patriknw:
Tested and with this it distributes the load over all (3) tservers.

Fixes #174

### DIFF
--- a/ddl-scripts/create_tables_yugabyte.sql
+++ b/ddl-scripts/create_tables_yugabyte.sql
@@ -22,7 +22,7 @@ CREATE TABLE IF NOT EXISTS event_journal(
 
 -- `event_journal_slice_idx` is only needed if the slice based queries are used
 CREATE INDEX IF NOT EXISTS event_journal_slice_idx ON event_journal(slice ASC, entity_type ASC, db_timestamp ASC)
-  INCLUDE (persistence_id, seq_nr, deleted);
+  SPLIT AT VALUES ((127), (255), (383), (511), (639), (767), (895));
 
 CREATE TABLE IF NOT EXISTS snapshot(
   slice INT NOT NULL,
@@ -56,7 +56,7 @@ CREATE TABLE IF NOT EXISTS durable_state (
 
 -- `durable_state_slice_idx` is only needed if the slice based queries are used
 CREATE INDEX IF NOT EXISTS durable_state_slice_idx ON durable_state(slice ASC, entity_type ASC, db_timestamp ASC)
-  INCLUDE (persistence_id, revision);
+  SPLIT AT VALUES ((127), (255), (383), (511), (639), (767), (895));
 
 -- Primitive offset types are stored in this table.
 -- If only timestamp based offsets are used this table is optional.


ddl-scripts/create_tables_yugabyte.sql
```diff
@@ -22,7 +22,7 @@ CREATE TABLE IF NOT EXISTS event_journal(
 
 -- `event_journal_slice_idx` is only needed if the slice based queries are used
 CREATE INDEX IF NOT EXISTS event_journal_slice_idx ON event_journal(slice ASC, entity_type ASC, db_timestamp ASC)
-  INCLUDE (persistence_id, seq_nr, deleted);
```
patriknw, line 25: Seems like it's not possible to combine INCLUDE and SPLIT AT, but I don't think we benefit much from the INCLUDE anyway. One thing we can watch out for is the `deleted = false` condition.

```
ERROR:  Illegal state: Number of split row values must be equal to number of primary key columns: 4 vs. 3
```

patriknw: the original reason for using the INCLUDE was for the backtracking queries that don't fetch the event payload

I created issue for follow up https://github.com/akka/akka-persistence-r2dbc/issues/184

johanandren: How does this work when there are three keys in the index, the docs only show a sample of a single primary key not using the full value-space. Is splitting at the hash value for all three combined somehow?

patriknw: `SPLIT AT VALUES` is used for range based pk/idx, i.e. when using `ASC` not `HASH`. Here we split it on the first column `slice`. There is some syntax to be able to split it on combined column values, see https://docs.yugabyte.com/latest/api/ysql/the-sql-language/statements/ddl_create_table/#split-at-values

johanandren: Ah, thx I looked at the wrong docs section, so it can actually be a tuple up to the number of key parts. Got it!

johanandren (approved):

ddl-scripts/create_tables_yugabyte.sql
patriknw: yes, it's not really documented for index, but seems to work the same way as for tables, which makes sense since an index is pretty much the same as an additional table

PR #190, opened 2022-01-24
Split projection_timestamp_offset table

patriknw:


ddl-scripts/create_tables_yugabyte.sql
```diff
@@ -85,7 +85,7 @@ CREATE TABLE IF NOT EXISTS akka_projection_timestamp_offset_store (
   -- the consumer lag is timestamp_consumed - timestamp_offset
   timestamp_consumed timestamp with time zone NOT NULL,
   PRIMARY KEY(slice ASC, projection_name ASC, timestamp_offset ASC, persistence_id ASC, seq_nr ASC)
-);
+) SPLIT AT VALUES ((127), (255), (383), (511), (639), (767), (895));
```
octonato, line 88: what does it do exactly?

octonato (approved):
Looks good, but I have a question...

No blocking, so approving.  :-)

ddl-scripts/create_tables_yugabyte.sql
patriknw: This table has a range (and not hash) primary key. Defined with the ASC keyword. By default that will start with only 1 tablet. This will pre-split the table into 8 tablets based on the slice column, which can have values 0 - 1023.

It's the same kind of tablet splitting as we already have for the journal and durable_state slice index.

PR #192, opened 2022-01-26
Update to R2DBC 0.9.0

patriknw:
* stmt .add() must not be at the end, any more
* workaround for connection pool bug when maxLifeTime isn't set
* verified that performance is the same

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/ConnectionFactoryProvider.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/ConnectionFactoryProvider.scala
@@ -95,6 +95,8 @@ class ConnectionFactoryProvider(system: ActorSystem[_]) extends Extension {
       .maxAcquireTime(JDuration.ofMillis(settings.acquireTimeout.toMillis))
       .acquireRetry(settings.acquireRetry)
       .maxIdleTime(JDuration.ofMillis(settings.maxIdleTime.toMillis))
+      // setting lifetime due to issue https://github.com/r2dbc/r2dbc-pool/issues/129
+      .maxLifeTime(JDuration.ofDays(365 * 100))
 
     if (settings.validationQuery.nonEmpty)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/akka/persistence/r2dbc/ConnectionFactoryProvider.scala
```diff
@@ -95,6 +95,8 @@ class ConnectionFactoryProvider(system: ActorSystem[_]) extends Extension {
       .maxAcquireTime(JDuration.ofMillis(settings.acquireTimeout.toMillis))
       .acquireRetry(settings.acquireRetry)
       .maxIdleTime(JDuration.ofMillis(settings.maxIdleTime.toMillis))
+      // setting lifetime due to issue https://github.com/r2dbc/r2dbc-pool/issues/129
+      .maxLifeTime(JDuration.ofDays(365 * 100))
```
patriknw, line 99: workaround for https://github.com/r2dbc/r2dbc-pool/issues/129
found this when verifying performance, connections are closed immediately otherwise

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -205,7 +205,8 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
       val result = r2dbcExecutor.updateInBatchReturning(s"batch insert [$persistenceId], [$totalEvents] events")(
         connection =>
           events.foldLeft(connection.createStatement(insertSql)) { (stmt, write) =>
-            bind(stmt, write).add()
+            stmt.add()
+            bind(stmt, write)
```
patriknw, line 209: this was because of the change in https://github.com/pgjdbc/r2dbc-postgresql/issues/463

johanandren (approved):

PR #195, opened 2022-01-27
URLEncoder.encode for jdk 8

patriknw:
https://github.com/akka/akka-persistence-r2dbc/runs/4962877931?check_suite_focus=true#step:4:47


johanandren (approved):
LGTM (but also wonder if it is important that we support JDK 8 given that it is already past EOL)

PR #203, opened 2022-02-07
Projection magement clear/set offset

patriknw:
It wasn't fully implemented for timestamp offsets.

### DIFF
--- a/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -7,6 +7,7 @@ package akka.projection.r2dbc.internal
 import java.time.Clock
 import java.time.Instant
 import java.time.{ Duration => JDuration }
+import java.util.UUID
 import java.util.concurrent.atomic.AtomicBoolean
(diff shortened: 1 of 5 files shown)


projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -467,21 +453,7 @@ private[projection] class R2dbcOffsetStore(
...
-        } else {
-          // TODO Try Batch without bind parameters for better performance. Risk of sql injection for these parameters is low.
-          val boundStatement =
-            filteredRecords.foldLeft(statement) { (stmt, rec) =>
-              stmt.add()
-              bindRecord(stmt, rec)
-            }
-          R2dbcExecutor.updateBatchInTx(boundStatement)
-        }
+      val offsetInserts = insertTimestampOffsetInTx(conn, filteredRecords)
```
patriknw, line 456: extracted this part into `def insertTimestampOffsetInTx` because same is needed from the `managementSetOffset`

johanandren (approved):
LGTM if the TODO/FIXMEs shouldn't be done in this PR

patriknw:
the FIXME, TODO already existed in previous code and they are not critical

Issue #82, opened 2021-10-26
EventsByTagQuery


patriknw:
eventsByTag and currentEventsByTag

Same offset tracking as for eventsBySlices.

PR #206, opened 2022-02-17
Store tags, but no query yet

patriknw:
* Nice to not just ignore the tags but at least store them for possible future usage.
* I think it can be useful to have a eventsBySlices query with additional tag condition
  to be able to have a filter of what entities to fetch.

Table changes:
```
ALTER TABLE event_journal ADD tags TEXT ARRAY;
ALTER TABLE durable_state ADD tags TEXT ARRAY;
```

See #82


patriknw:
btw, this thought came from selective event replication spike I was working on

johanandren (approved):
LGTM

octonato (approved):

PR #207, opened 2022-02-17
make it possible to configure a connection from url

octonato:


CONTRIBUTING.md
```diff
@@ -14,7 +14,7 @@ docker-compose -f docker/docker-compose-postgres.yml up
 docker exec -i docker_postgres-db_1 psql -U postgres -t < ddl-scripts/create_tables_postgres.sql
 ```
 
-##3 Some useful debug queries for Postgres
+## Some useful debug queries for Postgres
```
octonato, line 17: some small typo fixes

core/src/test/scala/akka/persistence/r2dbc/R2dbcSettingsSpec.scala
```diff
@@ -18,6 +18,21 @@ class R2dbcSettingsSpec extends AnyWordSpec with TestSuite with Matchers {
...
+
+    "support connection settings build from url" in {
+      val config =
+        ConfigFactory
+          .parseString("akka.persistence.r2dbc.connection-factory.url=whatever-url")
+          .withFallback(ConfigFactory.load())
+
+      val settings = R2dbcSettings(config.getConfig("akka.persistence.r2dbc"))
+      settings.connectionFactorySettings shouldBe a[ConnectionFactorySettings]
+      settings.connectionFactorySettings.urlOption shouldBe defined
```
octonato, line 35: The tests is really minimal. It only check the config object. 

We can move some DB tests to url based config eventually.

johanandren (approved):

core/src/main/scala/akka/persistence/r2dbc/R2dbcSettings.scala
```diff
@@ -82,6 +82,11 @@ final class QuerySettings(config: Config) {
  */
 @InternalStableApi
 final class ConnectionFactorySettings(config: Config) {
+
+  val urlOption: Option[String] =
+    if (config.hasPath("url")) Option(config.getString("url"))
+    else None
```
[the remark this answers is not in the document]
patriknw: We don't use `null` config values in Akka, but instead `""` for an optional property like this. It would still be an Option in code. I don't have a strong opinion about that. Maybe ConfigChecker (finding typos) doesn't work with null values, I don't know.

core/src/main/resources/reference.conf
```diff
@@ -100,6 +100,11 @@ akka.persistence.r2dbc {
 
   connection-factory {
     driver = "postgres"
+
+    # the connection can be configured with a url, eg: "r2dbc:postgresql://<host>:5432/<database>"
+    url = null
```
patriknw, line 105: I don't mind adding this, but curious about if there is any background to why we need an alternative way?

patriknw (approved):
LGTM

core/src/main/resources/reference.conf
octonato: True, actually I took this path because I want to add `DB_CLOSE_DELAY=-1` to the URL. Which is a source of silly mistakes when using h2 in tests. 

We can pass arbitrary options though, but I thought that it would be better to have one simple way of configured a in-mem db instead of enriching the settings with extra things.

octonato:
> We don't use null config values in Akka, but instead "" for an optional property like this. It would still be an Option in code. I don't have a strong opinion about that. Maybe ConfigChecker (finding typos) doesn't work with null values, I don't know.

I think it will be better to use "" indeed then I can simply filter on the option.

PR #209, opened 2022-02-18
add executeDdls for several ddl statements in same tx

patriknw:
@johanandren I noticed you could benefit from this, and we should probably use it in other places where we create table and index


johanandren (approved):
LGTM

PR #210, opened 2022-02-24
Upper bound on db_timestamp in eventsBySlices query, #178

### LINKED ISSUE #178: Investigate why eventsBySlices query has long response time when far from the tail
As can be seen in `explain analyze` results it is slow when querying from an offset far behind latest.
* https://github.com/akka/akka-persistence-r2dbc/issues/175#issuecomment-993410245
* https://github.com/akka/akka-persistence-r2dbc/issues/65#issuecomment-958069259

Is it not using the LIMIT as we would expect? Is it performing a distributed join with the parent table for "all" rows and not only those matching within the LIMIT?

patriknw:
* reduce result set that is sorted before limit is applied
* fetch event counts in buckets with aggregate query and use that
  to estimate the `db_timestamp < ?` that is needed to reach the limit
* also include some more columns in index to make the backtracking
  query IndexOnly Scan (no fetch from parent table)

Fixes #178

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
@@ -4,9 +4,11 @@
 
 package akka.persistence.r2dbc.internal
 
+import scala.collection.immutable
 import java.time.Instant
 import java.time.{ Duration => JDuration }
 
+import scala.annotation.tailrec
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration.Duration
@@ -29,7 +31,7 @@ import org.slf4j.Logger
 
   object QueryState {
     val empty: QueryState =
-      QueryState(TimestampOffset.Zero, 0, 0, 0, backtracking = false, TimestampOffset.Zero)
+      QueryState(TimestampOffset.Zero, 0, 0, 0, backtracking = false, TimestampOffset.Zero, Buckets.empty)
   }
 
   final case class QueryState(
@@ -38,7 +40,8 @@ import org.slf4j.Logger
       queryCount: Long,
       idleCount: Long,
       backtracking: Boolean,
-      latestBacktracking: TimestampOffset) {
+      latestBacktracking: TimestampOffset,
+      buckets: Buckets) {
 
     def currentOffset: TimestampOffset =
       if (backtracking) latestBacktracking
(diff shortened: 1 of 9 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -71,6 +152,13 @@ import org.slf4j.Logger
         toTimestamp: Option[Instant],
         behindCurrentTime: FiniteDuration,
         backtracking: Boolean): Source[SerializedRow, NotUsed]
+
+    def eventCountBuckets(
+        entityType: String,
+        minSlice: Int,
+        maxSlice: Int,
+        fromTimestamp: Instant,
+        limit: Int): Future[Seq[(Long, Long)]]
```
johanandren, line 161: Maybe do a case class for this vs tuple long long for clarity?

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/DurableStateDao.scala
```diff
@@ -336,4 +336,13 @@ private[r2dbc] class DurableStateDao(settings: R2dbcSettings, connectionFactory:
     Source.futureSource(result.map(Source(_))).mapMaterializedValue(_ => NotUsed)
   }
 
+  override def eventCountBuckets(
+      entityType: String,
+      minSlice: Int,
+      maxSlice: Int,
+      fromTimestamp: Instant,
+      limit: Int): Future[Seq[(Long, Long)]] = {
+    // FIXME look into if this is also needed for durable state
```
johanandren, line 345: Separate issue?

ddl-scripts/create_tables_postgres.sql
```diff
@@ -1,4 +1,4 @@
-CREATE TABLE IF NOT EXISTS event_journal(
+CREATE TABLE IF NOT EXISTS test_journal(
```
johanandren, line 1: Probably not intentional change, or else index needs update as well?

johanandren (approved):
LGTM with some small comments

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -102,20 +191,29 @@ import org.slf4j.Logger
...
       state.copy(latest = extractOffset(envelope), rowCount = state.rowCount + 1)
 
-    def nextQuery(state: QueryState, toDbTimestamp: Instant): (QueryState, Option[Source[Envelope, NotUsed]]) = {
-      // FIXME why is this rowCount -1 of expected?, see test EventsBySliceSpec "read in chunks"
-      if (state.queryCount == 0L || state.rowCount >= settings.querySettings.bufferSize - 1) {
+    def nextQuery(state: QueryState, endTimestamp: Instant): (QueryState, Option[Source[Envelope, NotUsed]]) = {
+      // Note that we can't know how many events with the same timestamp that are filtered out
+      // so continue until rowCount is 0. That means an extra query at the end to make sure there are no
+      // more to fetch.
+      if (state.queryCount == 0L || state.rowCount > 0) {
```
patriknw, line 198: this was an "unrelated" bug: https://github.com/akka/akka-persistence-r2dbc/issues/211

ddl-scripts/create_tables_postgres.sql
patriknw: thanks, yes, leftover from my testing

johanandren (approved):

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/DurableStateDao.scala
patriknw: https://github.com/akka/akka-persistence-r2dbc/issues/212

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
patriknw: done

PR #220, opened 2022-03-07
Support ssl-mode values, in addition to enum names

patriknw:


core/src/main/scala/akka/persistence/r2dbc/ConnectionFactoryProvider.scala
```diff
@@ -84,7 +84,7 @@ class ConnectionFactoryProvider(system: ActorSystem[_]) extends Extension {
       builder.option(ConnectionFactoryOptions.SSL, java.lang.Boolean.TRUE)
 
       if (settings.sslMode.nonEmpty)
-        builder.option(PostgresqlConnectionFactoryProvider.SSL_MODE, SSLMode.valueOf(settings.sslMode))
+        builder.option(PostgresqlConnectionFactoryProvider.SSL_MODE, SSLMode.fromValue(settings.sslMode))
```
patriknw, line 87: this does the right thing

raboof (approved):

patriknw:
yes, it works with verify-ca as well

marcoderama:
D'oh!  I see `test` in the path now...  ([core/src/test/scala/akka/persistence/r2dbc/R2dbcSettingsSpec.scala](https://github.com/akka/akka-persistence-r2dbc/pull/220/files#diff-8715b5ab2df63d203a2a6aa2bb274ebb4e50f50f25eebdd76e738b0ba366c240))  Thanks.

PR #221, opened 2022-03-07
Update documentation of config for durable state store

patriknw:


raboof (approved):
Matches https://github.com/akka/akka-persistence-r2dbc/blob/main/core/src/main/resources/reference.conf#L43

'state' seems a bit generic but I guess 'durable state' isn't much better anyway, good to have the docs matching reality ;)

PR #226, opened 2022-04-01
Upper bound on db_timestamp in changesBySlices query, #212

### LINKED ISSUE #212: Add upper timestamp bound on eventsBySlices for DurableState
I think the same problem as https://github.com/akka/akka-persistence-r2dbc/issues/178 exists for DurableState. Should be fairly easy to implement in the same way (much of that code is shared).

patriknw:
Corresponding to https://github.com/akka/akka-persistence-r2dbc/pull/210 but for Durable State.

Fixes #212


pvlugter (approved):
LGTM

PR #237, opened 2022-05-10
Set max-life-time to shrink pool

patriknw:
* Even though idle connections are evicted from the pool it's
  likeley that connections in the pool are "touched" also during infrequent usage,
  i.e. they are not idle.
* By setting the max-life-time the connections are cycled out from the pool and
  the number of connections in the pool will shrink, as long as there is not demand
  for more connections.

I tested this with R2dbcJournalPerfManyActorsSpec by adding a "slow" phase at the end. Also added some temporary debug logging in r2dbc-pool `evictionPredicate`.

Number of connections can be seen with:
```
SELECT count(*)
FROM   pg_stat_activity
WHERE  datname = current_database()  -- only current database
AND    pid <> pg_backend_pid()
```

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -141,8 +141,14 @@ akka.persistence.r2dbc {
     acquire-retry = 20
 
     # Maximum idle time of the connection in the pool.
-    # This value is used as an interval for background eviction of idle connections.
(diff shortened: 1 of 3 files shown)


jrudolph (approved):
LGTM

PR #238, opened 2022-05-13
Fix connection leak caused by create-timeout, #182

### LINKED ISSUE #182: Connection leak
When running the benchmarks at high load there seems to be more connections than the configured max pool size.

Ran 8 pods with max 50 each.

```
SELECT *
FROM   pg_stat_activity
WHERE  datname = current_database()  -- only current database
AND    pid <> pg_backend_pid()       -- except your current session
AND    state LIKE 'idle%';
```

returns 296 rows (and that is for one of the tservers)

patriknw:
* createTimeout is just a timeout on the Mono, but the underlying connection may still be established
  and not closed
* use ConnectionFactoryOptions.CONNECT_TIMEOUT instead
* also change acquire-retry to 1, and mention that it can be increased

Fixes #182

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -132,13 +132,14 @@ akka.persistence.r2dbc {
     # Maximum pool size.
     max-size = 20
     # Maximum time to create a new connection.
-    create-timeout = 3 seconds
+    connect-timeout = 3 seconds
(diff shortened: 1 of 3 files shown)


raboof (approved):
great find!

patriknw:
I'll report upstream that createConnection is dangerous.

PR #239, opened 2022-05-19
Add support for using json/jsonb as column type for payloads (pg only)

odd:
* All test data has been changed to be valid json (test inheriting from plugin tck classes which uses non json compatible payloads have been marked as ignored for now) 
* New settings added:
  ```
  akka.persistence.r2dbc.journal.payload-column-type = "json" | "jsonb" | "bytea"
  akka.persistence.r2dbc.snapshot.payload-column-type = "json" | "jsonb" | "bytea"
  akka.persistence.r2dbc.state.payload-column-type = "json" | "jsonb" | "bytea"
  ```

* Create the database using the jsonb ddl script variant at:
  ```
  ddl-scripts/create_tables_postgres_jsonb.sql
  ```
* Run tests with:
    ```
    sbt -Dakka.persistence.r2dbc.journal.payload-column-type=jsonb  -Dakka.persistence.r2dbc.snapshot.payload-column-type=jsonb  -Dakka.persistence.r2dbc.state.payload-column-type=jsonb test
```

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -7,6 +7,9 @@ akka.persistence.r2dbc {
     # name of the table to use for events
     table = "event_journal"
 
+    # the column type to use for event payloads (bytea or jsonb)
+    payload-column-type = "BYTEA"
+
     # Otherwise it would be a pinned dispatcher, see https://github.com/akka/akka/issues/31058
     plugin-dispatcher = "akka.actor.default-dispatcher"
 
@@ -58,6 +61,9 @@ akka.persistence.r2dbc {
     class = "akka.persistence.r2dbc.snapshot.R2dbcSnapshotStore"
     table = "snapshot"
 
+    # the column type to use for snapshot payloads (bytea or jsonb)
+    payload-column-type = "BYTEA"
+
     # Otherwise it would be a pinned dispatcher, see https://github.com/akka/akka/issues/31058
     plugin-dispatcher = "akka.actor.default-dispatcher"
 
@@ -74,6 +80,9 @@ akka.persistence.r2dbc {
     class = "akka.persistence.r2dbc.state.R2dbcDurableStateStoreProvider"
 
     table = "durable_state"
+
+    # the column type to use for durable state payloads (bytea or jsonb)
+    payload-column-type = "BYTEA"
 
     # When this is enabled the updates verifies that the revision is +1 of
     # previous revision. There might be a small performance gain if
--- a/core/src/test/scala/akka/persistence/r2dbc/journal/PersistTimestampSpec.scala
+++ b/core/src/test/scala/akka/persistence/r2dbc/journal/PersistTimestampSpec.scala
@@ -12,6 +12,8 @@ import akka.Done
 import akka.actor.testkit.typed.scaladsl.LogCapturing
 import akka.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit
 import akka.actor.typed.ActorSystem
+import akka.persistence.r2dbc.internal.PayloadCodec
+import akka.persistence.r2dbc.internal.PayloadCodec.RichRow
 import akka.persistence.r2dbc.R2dbcSettings
 import akka.persistence.r2dbc.TestActors.Persister
 import akka.persistence.r2dbc.TestConfig
@@ -31,7 +33,7 @@ class PersistTimestampSpec
   override def typedSystem: ActorSystem[_] = system
   private val settings = new R2dbcSettings(system.settings.config.getConfig("akka.persistence.r2dbc"))
   private val serialization = SerializationExtension(system)
-
+  private implicit val journalPayloadCodec: PayloadCodec = settings.journalPayloadCodec
   case class Row(pid: String, seqNr: Long, dbTimestamp: Instant, event: String)
 
   "Persist timestamp" should {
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: core/src/main/scala/akka/persistence/r2dbc/PayloadCodec.scala [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+import io.r2dbc.spi.Statement
+
+sealed trait PayloadCodec {
+  def encode(bytes: Array[Byte]): AnyRef
+  def decode(payload: AnyRef): Array[Byte]
+}
+object PayloadCodec {
+  case object ByteArrayCodec extends PayloadCodec {
+    override def encode(bytes: Array[Byte]): Array[Byte] = bytes
+    override def decode(payload: AnyRef): Array[Byte] = payload.asInstanceOf[ByteBuffer].array()
```
patriknw, line 19: previously this was
```
row.get(name, classOf[Array[Byte]])
```
but now ByteBuffer via
```
row.get(name)
```

I think I would prefer to keep the explicit byte array and let the driver retrieve that in the way it decides is most efficient.

core/src/test/scala/akka/persistence/r2dbc/journal/R2dbcJournalPerfManyActorsSpec.scala
```diff
@@ -17,6 +17,7 @@ import akka.persistence.journal.JournalPerfSpec.ResetCounter
 import akka.persistence.r2dbc.TestDbLifecycle
 import akka.testkit.TestProbe
 
+@org.scalatest.Ignore // TODO: Remove once data can be made json-compatible
```
patriknw, line 20: It's quite intrusive to make all test data json (or that kind of quoted strings). Can't we test this feature in some other way? Some dedicated test that is verifying that json can be stored and retrieved for the journal, snapshot and read from queries, instead of aiming for running all tests with json?

patriknw (commented):
This is a useful addition, thanks for contributing @odd

core/src/main/resources/reference.conf
```diff
@@ -7,6 +7,9 @@ akka.persistence.r2dbc {
     # name of the table to use for events
     table = "event_journal"
 
+    # the column type to use for event payloads (bytea or jsonb)
+    payload-column-type = "bytea"
```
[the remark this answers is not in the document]
odd: Fixed

core/src/main/scala/akka/persistence/r2dbc/PayloadCodec.scala
```diff
@@ -0,0 +1,31 @@
...
+import java.nio.ByteBuffer
+import io.r2dbc.postgresql.codec.Json
+import io.r2dbc.spi.Row
+import io.r2dbc.spi.Statement
+
+sealed trait PayloadCodec {
+  def encode(bytes: Array[Byte]): AnyRef
+  def decode(payload: AnyRef): Array[Byte]
+}
+object PayloadCodec {
```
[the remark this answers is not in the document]
odd: Fixed

core/src/main/scala/akka/persistence/r2dbc/PayloadCodec.scala
odd: Fixed

core/src/test/scala/akka/persistence/r2dbc/journal/R2dbcJournalPerfManyActorsSpec.scala
odd: Yes, I feel the same way. I did it this way to make sure I didn't miss any functionality where changing the payload column type might have an impact (I haven't worked on this codebase until this PR). But now that all tests pass (except the intentionally ignored), changing to specialized test (and reverting the test data changes) are probably a better idea. I'll start working on that and get back once I have something in place. Thanks for the quick feedback!

patriknw:
@odd I see that I haven't payed attention to the updates here, sorry. I'll take a look next week.

core/src/test/scala/akka/persistence/r2dbc/journal/PersistTimestampSpec.scala
```diff
@@ -48,12 +50,12 @@ class PersistTimestampSpec
       (1 to 100).foreach { n =>
         val p = n % numberOfEntities
         // mix some persist 1 and persist 3 events
+        val event = s"e$p-$n"
```
patriknw, line 53: ok, so an ordinary String without any json formatting can also be stored in `JSONB`, and that's why all tests can be run in that mode?

patriknw (commented):
looking good

patriknw:
Missing documentation but otherwise ready I guess?

core/src/test/scala/akka/persistence/r2dbc/journal/PersistTimestampSpec.scala
odd: No, the test will fail if you use the JSON or JSONB schema while running the tests (that's why I made all the test data be valid JSON in my original request). I have been planning on adding separate tests that run using JSON data stored in a JSON schema, but haven't gotten around to it yet unfortunately. I will try to put something together before the end of the summer.

patriknw: I see. I thought it just worked because you still had the ci workflow with ` -Dakka.persistence.r2dbc.journal.payload-column-type=jsonb`. Makes sense to have a separate test.

patriknw:
> Missing documentation but otherwise ready I guess?

@odd any chance you have some cycles to complete this or do you want us to merge it and we can add the documentation?

odd:
Hi @patriknw, sorry for the late reply. I am unfortunately too busy at work at the moment any work on the documentation/tests from my side will have to wait until later this spring. You are very welcome to merge it and complete the work on your side.

patriknw:
Thank you @odd

odd:
You’re welcome, happy to be of service!

PR #257, opened 2022-06-22
Allow envelope from backtracking as duplicate when in-flight, #245

### LINKED ISSUE #245: Rejected envelope from backtracking
```
[2022-06-02 10:01:40,705] [WARN] [akka.projection.r2dbc.internal.R2dbcOffsetStore] [] [] [ShoppingAnalyticsService-akka.actor.default-dispatcher-16] - Rejecting unexpected sequence number [1] for pid [ShoppingCart|20220602-081454.1654150494-115], previous sequence number [7]. Offset: TimestampOffset(2022-06-02T06:14:54.854098Z,1970-01-01T00:00:00Z,Map(ShoppingCart|20220602-081454.1654150494-115 -> 1))
[2022-06-02 10:01:40,711] [ERROR] [akka.projection.r2dbc.internal.R2dbcProjectionImpl$R2dbcInternalProjectionState] [akka://ShoppingAnalyticsService@127.0.0.1:3551] [] [ShoppingAnalyticsService-akka.actor.default-dispatcher-3] - [cart-events-ShoppingCart-768-1023] Failed to process envelope with offset [TimestampOffset(2022-06-02T06:14:54.854098Z,1970-01-01T00:00:00Z,Map(ShoppingCart|20220602-081454.1654150494-115 -> 1))]. Projection will stop as defined by recovery strategy.
java.lang.IllegalStateException: Rejected envelope from backtracking, persistenceId [ShoppingCart|20220602-081454.1654150494-115], seqNr [1] due to unexpected sequence number. Please report this issue at https://github.com/akka/akka-persistence-r2dbc
	at akka.projection.r2dbc.internal.R2dbcOffsetStore.isAccepted(R2dbcOffsetStore.scala:673)
	at akka.projection.r2dbc.internal.R2dbcOffsetStore.isAccepted(R2dbcOffsetStore.scala:591)
```

Looking at the code in `isAccepted` it passes the `isDuplicate` check so maybe related to `currentInflight`. It's an `atLeastOnceAsync` projection.

patriknw:
This was noticed as a problem with atLeastOnceFlow but could probably happen with other async projections as well.

The scenario to trigger this was that events were added to in-flight but offset not stored yet. Then a backtracking query emitted previous events that were rejected (throw exception) because seqNr wasn't the expected previous+1. Solution is to accept such events as duplicates when they are in-flight.

Fixes #245

### DIFF
--- a/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -526,6 +526,11 @@ private[projection] class R2dbcOffsetStore(
           }
         case _ => true
       }
+    if (newInflight.size >= 10000) {
+      throw new IllegalStateException(
+        s"Too many envelopes in-flight [${newInflight.size}]. " +
+        "Please report this issue at https://github.com/akka/akka-persistence-r2dbc")
+    }
     if (!inflight.compareAndSet(currentInflight, newInflight))
       cleanupInflight(newState) // CAS retry, concurrent update of inflight
   }
@@ -671,6 +676,10 @@ private[projection] class R2dbcOffsetStore(
         val ok = seqNr == prevSeqNr + 1
         if (ok) {
           FutureTrue
+        } else if (seqNr <= currentInflight.getOrElse(pid, 0L)) {
+          // currentInFlight contains those that have been processed or about to be processed in Flow,
+          // but offset not saved yet => ok to handle as duplicate
+          FutureFalse
         } else if (recordWithOffset.envelopeLoaded) {
           logUnexpected()
           FutureFalse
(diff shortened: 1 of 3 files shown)


projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -526,6 +526,11 @@ private[projection] class R2dbcOffsetStore(
           }
         case _ => true
       }
+    if (newInflight.size >= 10000) {
```
patriknw, line 529: added this as a precaution of any bugs resulting in inFlight not cleaned up as expected

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -671,6 +676,10 @@ private[projection] class R2dbcOffsetStore(
         val ok = seqNr == prevSeqNr + 1
         if (ok) {
           FutureTrue
+        } else if (seqNr <= currentInflight.getOrElse(pid, 0L)) {
```
patriknw, line 679: yeah, hard work for a single line fix

johanandren (approved):

PR #258, opened 2022-06-22
Tagged pubsub

leviramsey:
#206 partially implements support for tagging events by persisting them (without adding query support see #82).

When tagged events are persisted, the `Tagged` wrapper is unwrapped, but this does not occur when the events are broadcast via pubsub.  The result is that queries will not see the `Tagged` wrapper when polling the DB but will when they receive events via pubsub.

This change unwraps the tagged events in pubsub.  This approach can be changed in the future, e.g. if a hypothetical `eventsByTagAndSlice` query is introduced.

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/PubSub.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/PubSub.scala
@@ -17,6 +17,7 @@ import akka.actor.typed.pubsub.Topic
 import akka.annotation.InternalApi
 import akka.persistence.Persistence
 import akka.persistence.PersistentRepr
+import akka.persistence.journal.Tagged
 import akka.persistence.query.TimestampOffset
 import akka.persistence.query.typed.EventEnvelope
 import akka.persistence.typed.PersistenceId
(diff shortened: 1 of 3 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/PubSub.scala
```diff
@@ -55,11 +56,27 @@ import akka.persistence.typed.PersistenceId
     val slice = persistenceExt.sliceForPersistenceId(pid)
 
     val offset = TimestampOffset(timestamp, timestamp, Map(pid -> pr.sequenceNr))
+    val notTagged =
+      pr.payload match {
+        case Tagged(payload, _) =>
+          // eventsByTag not implemented (see issue #82), but events can still be tagged, so we unwrap this tagged event.
+          PersistentRepr(
```
patriknw, line 63: use this instead: `pr.withPayload(payload)`

patriknw: but this copy of the `PersistentRepr` isn't needed at all. This match can extract the `Tagged(payload, _)` or `pr.payload` and then when creating the EventEnvelope use that extracted payload instead of `Option(pr.payload)`

core/src/main/scala/akka/persistence/r2dbc/journal/R2dbcJournal.scala
```diff
@@ -187,7 +187,9 @@ private[r2dbc] final class R2dbcJournal(config: Config, cfgPath: String) extends
     if (pubSub.isDefined) {
```
patriknw, line 187: while you are at it, could you change this to:
```
  match pubSub {
    case Some(p) =>
       dbTimestamp.map { timestamp =>
         messages.iterator
             .flatMap(_.payload.iterator)
             .foreach(pr => p.publish(pr, timestamp))
         Done
       }
    case None =>
      dbTimestamp.map(_ => Done)(ExecutionContexts.parasitic)
  }
```

patriknw (commented):
looking good, just minor suggestion

patriknw (approved):
LGTM, thanks

PR #265, opened 2022-07-07
Customizable entity type extraction in migration

leviramsey:
The migration tool uses `PersistenceId.extractEntityType` to derive an entity type from a persistence ID.  That method hardcodes the convention used in Lagom's Scala DSL, which isn't universal (Lagom's Java DSL doesn't follow it and non-Lagom users may have some other convention for encoding entity type in the persistence ID).

This change allows a custom entity-type extractor to be injected as a constructor parameter, defaulting to `PersistenceId.extractEntityType` for consistency with prior behavior.

Admittedly, I'm not entirely sure of the significance of the entity type in the journal schema, but I assume it's there for a reason (e.g. facilitating an `eventsByEntityTypeAndSlice` or similar query), so it's better to have it than not have it.

### DIFF
--- a/migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
+++ b/migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
@@ -93,7 +93,7 @@ object MigrationTool {
  *
  * Note: tags are not migrated.
  */
-class MigrationTool(system: ActorSystem[_]) {
+class MigrationTool(system: ActorSystem[_], entityTypeExtractor: String => String = PersistenceId.extractEntityType) {
   import MigrationTool.Result
   import system.executionContext
   private implicit val sys: ActorSystem[_] = system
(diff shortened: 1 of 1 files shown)


migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
```diff
@@ -93,7 +93,7 @@ object MigrationTool {
  *
  * Note: tags are not migrated.
  */
-class MigrationTool(system: ActorSystem[_]) {
+class MigrationTool(system: ActorSystem[_], entityTypeExtractor: String => String = PersistenceId.extractEntityType) {
```
patriknw, line 96: I don't think this will help much, because `PersistenceId.extractEntityType` is used all over the place when persisting events and durable state. It's based on the assumption that the default `PersistenceId` from Akka Typed is used.

In general I think a migration tool like this should be seen as an example that can be adjusted to application specific needs by modifying the source code (beauty of OSS).

If we want to make the binary more flexible I think a better solution is to be able to define a mapping function for the EventEnvelope from the source (sourceEventsByPersistenceIdQuery). Then it would be possible to rewrite the persistenceId for example. That is already possible by writing your own `CurrentEventsByPersistenceIdQuery`, but that is maybe too complicated?

patriknw: entityType is used for eventsBySlices query, yes

patriknw:
closing due to inactivity and not clear why we need this

PR #278, opened 2022-09-23
Implement Delete effect for DurableState, #276

### LINKED ISSUE #276: Implement Delete effect for DurableState
https://github.com/akka/akka/pull/31529

patriknw:
* When deleting it will keep the record in the db, to keep track of latest revision number.
* Kept hard delete for revision=0 parameter.
* Emits DeletedDurableState in eventsBySlices query.

Fixes #276

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/R2dbcDurableStateStore.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/R2dbcDurableStateStore.scala
@@ -13,6 +13,7 @@ import akka.actor.ExtendedActorSystem
 import akka.actor.typed.scaladsl.LoggerOps
 import akka.actor.typed.scaladsl.adapter._
 import akka.persistence.Persistence
+import akka.persistence.query.DeletedDurableState
 import akka.persistence.query.DurableStateChange
 import akka.persistence.query.Offset
 import akka.persistence.query.TimestampOffset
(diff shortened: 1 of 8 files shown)


core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/R2dbcDurableStateStore.scala
```diff
@@ -93,22 +116,30 @@ class R2dbcDurableStateStore[A](system: ExtendedActorSystem, config: Config, cfg
...
     deleteObject(persistenceId, revision = 0)
 
+  /**
+   * Delete the value, which will fail with `IllegalStateException` if the existing stored `revision` + 1 isn't equal to
+   * the given `revision`. This optimistic locking check can be disabled with configuration `assert-single-writer`. The
+   * stored revision for the persistenceId is updated and next call to [[getObject]] will return the revision, but with
+   * no value.
+   *
+   * If the given revision is `0` it will fully delete the value and revision from the database without any optimistic
+   * locking check. Next call to [[getObject]] will then return revision 0 and no value.
```
octonato, line 139: Is that how we will give the option of a hard delete? Sound a bit hacky, no?

Wouldn't it be better to have an explicit method for the hard-delete path?

johanandren (approved):
LGTM (but may need a rebase/retrigger CI so we know it builds on Scala 2.12)

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/R2dbcDurableStateStore.scala
patriknw: We don't have any API for hard delete in Akka DurableStateBehavior, and I don't think we should add it. Resetting revision will break queries/projections. It can be useful for some kind of backoffice tool and then it will be on the store directly. I left the hard delete case mostly for compatibility with previous behavior when using the deprecated deleteObject method without revision parameter. However, I don't think it's completely hacky to use revision=0 because that is what the revision will be after the delete.

octonato: Ok I see. 

Thanks for the clarifications.

octonato (approved):

PR #279, opened 2022-09-26
chore: Make it build with Scala 2.12

johanandren:
Needed for bringing Akka Projection gRPC into Akka Projection.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -56,7 +56,7 @@ jobs:
         uses: coursier/cache-action@v5
       - name: Code style check and binary-compatibility check
         run: |-
-          sbt -jvm-opts .jvmopts-ci ++2.12.16 compile Test/compile
+          sbt -jvm-opts .jvmopts-ci ++2.12 compile Test/compile
 
   test-postgres:
     name: Run test with Postgres
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -33,6 +33,31 @@ jobs:
         run: |-
           sbt -jvm-opts .jvmopts-ci scalafmtCheckAll scalafmtSbtCheck headerCheck
 
+  compile-with-scala-2_12:
+    name: Compile with Scala 2.12
+    runs-on: ubuntu-18.04
+    if: github.repository == 'akka/akka-persistence-r2dbc'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # we don't know what commit the last tag was it's safer to get entire repo so previousStableVersion resolves
+          fetch-depth: 0
+      - name: Checkout GitHub merge
+        if: github.event.pull_request
+        run: |-
+          git fetch origin pull/${{ github.event.pull_request.number }}/merge:scratch
+          git checkout scratch
+      - name: Set up JDK 11
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.11.0-9
+      - name: Cache Coursier cache
+        uses: coursier/cache-action@v5
+      - name: Code style check and binary-compatibility check
+        run: |-
+          sbt -jvm-opts .jvmopts-ci ++2.12.16 compile Test/compile
+
   test-postgres:
     name: Run test with Postgres
     runs-on: ubuntu-18.04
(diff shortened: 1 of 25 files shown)


.github/workflows/build-test.yml
```diff
@@ -33,6 +33,31 @@ jobs:
...
+          git checkout scratch
+      - name: Set up JDK 11
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.11.0-9
+      - name: Cache Coursier cache
+        uses: coursier/cache-action@v5
+      - name: Code style check and binary-compatibility check
+        run: |-
+          sbt -jvm-opts .jvmopts-ci ++2.12.16 compile Test/compile
```
patriknw, line 59: we can use `++2.12.x`

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -58,10 +58,10 @@ import reactor.core.publisher.Mono
   def updateInTx(statements: immutable.IndexedSeq[Statement])(implicit
       ec: ExecutionContext): Future[immutable.IndexedSeq[Int]] =
     // connection not intended for concurrent calls, make sure statements are executed one at a time
-    statements.foldLeft(Future.successful(IndexedSeq.empty[Int])) { (acc, stmt) =>
+    statements.foldLeft(Future.successful(IndexedSeq.empty[Int].toIndexedSeq)) { (acc, stmt) =>
```
patriknw, line 61: That looks strange? would be better with `Vector.empty[Int]`

patriknw: or `immutable.IndexedSeq.empty`, but I would prefer Vector here

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -79,8 +79,8 @@ import reactor.core.publisher.Mono
       val consumer: BiConsumer[mutable.Builder[A, IndexedSeq[A]], A] = (builder, elem) => builder += elem
       Flux
         .from[A](result.map((row, _) => mapRow(row)))
-        .collect(() => IndexedSeq.newBuilder[A], consumer)
-        .map(_.result())
+        .collect[scala.collection.mutable.Builder[A, IndexedSeq[A]]](() => immutable.IndexedSeq.newBuilder[A], consumer)
```
patriknw, line 82: not important but for implementation details it could be `Vector.newBuilder`

projection/src/test/scala/akka/projection/r2dbc/EventSourcedPubSubSpec.scala
```diff
@@ -226,7 +225,7 @@ class EventSourcedPubSubSpec
             p.envelope.offset.asInstanceOf[TimestampOffset].timestamp == p.envelope.offset
               .asInstanceOf[TimestampOffset]
               .readTimestamp)
-        log.info("via pub-sub {}: {}", pid, viaPubSub.map(_.envelope.sequenceNr).mkString(", "))
+        log.infoN("via pub-sub {}: {}", pid, viaPubSub.map(_.envelope.sequenceNr).mkString(", "))
```
patriknw, line 228: note that we also have `info2`

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -263,7 +263,7 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger, logDb
 
         result.failed.foreach { exc =>
           if (log.isDebugEnabled())
-            log.debug("{} - DB call failed: {}", logPrefix, exc.toString)
+            log.debugN("{} - DB call failed: {}", logPrefix, exc.toString)
```
patriknw, line 266: `debug2`, might be more places

patriknw (commented):
a few nitpicks...

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
johanandren: I thought so as well and tried vector but couldn't get it to compile with the Java Flux/Mono method type params.

johanandren: Gave it another try but I'm giving up. Johan vs Scala 2.12 and Java lambda invariance: 0 - 1

PR #294, opened 2022-10-24
Update project status

patriknw:


project/project-info.conf
```diff
@@ -1,7 +1,7 @@
 project-info {
   version: "current"
   shared-info {
-    jdk-versions: ["OpenJDK 11"]
+    jdk-versions: ["OpenJDK 8", "OpenJDK 11"]
```
johanandren, line 4: But not 17 to align with Akka core?

johanandren (approved):
LGTM with one question

ennru (approved):
LGTM.

ennru:
The JavaDoc updates made the formatting invalid and scalafmt screams about it.

project/project-info.conf
patriknw: In general we should add CI jobs for 17, but for this one I'm fine with stating 17 immediately since we are running with 17.

patriknw: also, I'm not sure OpenJDK is the right name. We had Adopt in some others, but that is wrong, and I don't think we need to be so specific here.

johanandren (approved):

PR #300, opened 2022-11-02
Reduce periodic latency spikes

patriknw:
* With pub-sub enabled there are latency spikes when the db queries are running, especially the backtracking queries.
* Seems to help with mergePrioritized
* Also limited number of backtracking queries that will be run before mixing with ordinary query, for better fairness

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
@@ -31,7 +31,7 @@ import org.slf4j.Logger
 
   object QueryState {
     val empty: QueryState =
-      QueryState(TimestampOffset.Zero, 0, 0, 0, backtracking = false, TimestampOffset.Zero, Buckets.empty)
+      QueryState(TimestampOffset.Zero, 0, 0, 0, backtrackingCount = 0, TimestampOffset.Zero, Buckets.empty)
   }
 
   final case class QueryState(
@@ -39,9 +39,11 @@ import org.slf4j.Logger
       rowCount: Int,
       queryCount: Long,
       idleCount: Long,
-      backtracking: Boolean,
+      backtrackingCount: Int,
       latestBacktracking: TimestampOffset,
       buckets: Buckets) {
+
+    def backtracking: Boolean = backtrackingCount > 0
 
     def currentOffset: TimestampOffset =
       if (backtracking) latestBacktracking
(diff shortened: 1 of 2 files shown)


core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -167,7 +167,9 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
               pubSub.eventTopic(entityType, slice) ! Topic.Subscribe(ref.toTyped[EventEnvelope[Event]])
             }
           }
-      dbSource.merge(pubSubSource).via(deduplicate(settings.querySettings.deduplicateCapacity))
+      dbSource
+        .mergePrioritized(pubSubSource, leftPriority = 1, rightPriority = 10)
+        .via(deduplicate(settings.querySettings.deduplicateCapacity))
```
johanandren, line 172: Did look a bit at Merge trying to figure out if there is something wrt fairness, but nothing obviously unfair in there.

johanandren (commented):
Looks reasonable

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -331,7 +333,8 @@ import org.slf4j.Logger
     }
 
     def switchFromBacktracking(state: QueryState): Boolean = {
```
aludwiko, line 335: Interesting, is this backtracking somewhere documented? Or only in the code?

patriknw: Some is explained in https://github.com/akka/akka-persistence-r2dbc/blob/main/design-notes.md

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
patriknw: I tested some more locally and I indeed see that this helps, but I think the change that had biggest impact in the real testing was that I increased the number of projection instances from 4 to 16 (still running 4 ActorSystems). In that way there are less heavy bursts of events from the db queries.

I also played with the idea of adding `dbSource.throttle`. While that helps it would be difficult to find an optimal rate. Too low and there is a risk that the db queries fall behind the actual produced rate. Too high and it will not matter for fairness. Too low it will also suffer for the raw catch up throughput. No, wrong path.

johanandren (approved):

PR #313, opened 2022-11-17
Dynamically enable/disable publish-events

patriknw:
I'll try it out in a sample application for Projections over gRPC, but otherwise ready for review.


johanandren (approved):
LGTM

patriknw:
Thanks for review, marking as draft until I have tried it out some more.

patriknw:
I have tested this with the Projection over gRPC bench app. It looks good. In the end it would have to be an application specific testing/tuning option but I think we can have it enabled by default like this.

PR #318, opened 2022-11-22
ci: renovate GH workflows

ennru:
References https://github.com/akka/akka/pull/31730

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -14,24 +14,29 @@ jobs:
     if: github.repository == 'akka/akka-persistence-r2dbc'
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3.1.0
         with:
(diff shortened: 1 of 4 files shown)


johanandren (approved):

.github/workflows/build-test.yml
```diff
@@ -48,23 +53,27 @@ jobs:
...
-      - name: Set up JDK 11
-        uses: olafurpg/setup-scala@v10
-        with:
-          java-version: adopt@1.11.0-9
+
       - name: Cache Coursier cache
-        uses: coursier/cache-action@v5
+        uses: coursier/cache-action@v6.4.0
+
+      - name: Set up JDK 17
```
patriknw, line 60: this is setting up 11 not 17

patriknw: might be better to have this with 8 so that we validate that it's compiling with 8, and then tests can be run with 11 (and 17)

PR #319, opened 2022-11-23
ci: cross-test combinations in nightly tests

ennru:
References #318

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -42,7 +42,7 @@ jobs:
           sbt scalafmtCheckAll scalafmtSbtCheck headerCheck
 
   compile-with-scala-2_12:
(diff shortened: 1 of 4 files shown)


johanandren (approved):
Seems reasonable to me

.github/workflows/build-test.yml
```diff
@@ -104,7 +107,7 @@ jobs:
           sbt test
 
   test-yugabyte:
-    name: Run tests with Yugabyte
+    name: Run tests with Yugabyte on JDK 11
```
patriknw, line 110: Remember to update the branch protection rules to set these as required. For that reason I'm not convinced that it's worth having exact names on these, but not a strong opinion.

patriknw (approved):
LGTM

Issue #321, opened 2022-11-24
failed: R2dbcTimestampOffsetProjectionSpec


patriknw:
* Postgres, Scala 2.12, JDK 17
  * https://github.com/akka/akka-persistence-r2dbc/actions/runs/3537210990/jobs/5936976951#step:6:556
  * https://github.com/akka/akka-persistence-r2dbc/actions/runs/3533172986/jobs/5928459678#step:6:542
* Postgres, Scala 2.13, JDK 17
  * https://github.com/akka/akka-persistence-r2dbc/actions/runs/3537210990/jobs/5936976951#step:6:556
  * https://github.com/akka/akka-persistence-r2dbc/actions/runs/3533172986/jobs/5928459809#step:6:565

patriknw:
@ennru I understand what is causing the test failures when running JDK 17 on Linux. The precision has been expanded to nanoseconds. Previously it was microseconds. Postgres will only handle microseconds and then it's rounded and not matching the test assertions.

https://bugs.openjdk.org/browse/JDK-8242504

I don't see the same on Mac, because it's still microseconds on Mac.

I will come up with a solution.

PR #324, opened 2022-11-28
fix: Use explicit microsecond timestamp resolution, #321

patriknw:
* Truncate timestamps to microseconds because Postgres timestamps has the resolution of microseconds but some OS/JDK (Linux/JDK17) has Instant resolution of nanoseconds.
* This could have been a problem when sending the events directly vs reading through the db.
* Change TestClock to use microseconds or milliseconds and use it from tests instead of Instant.now().
* Some tests that are not using the timestamp offset must still use millisecond resolution.

References #321

### DIFF
--- a/projection/src/test/resources/logback-test.xml
+++ b/projection/src/test/resources/logback-test.xml
@@ -12,8 +12,8 @@
         <appender-ref ref="STDOUT"/>
     </logger>
 
-    <logger name="akka.projection.r2dbc" level="TRACE" />
-    <logger name="akka.persistence.r2dbc" level="TRACE" />
+    <logger name="akka.projection.r2dbc" level="DEBUG" />
+    <logger name="akka.persistence.r2dbc" level="DEBUG" />
 <!--    <logger name="io.r2dbc.postgresql.QUERY" level="DEBUG" />-->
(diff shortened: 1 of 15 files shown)


projection/src/test/resources/logback-test.xml
```diff
@@ -12,8 +12,8 @@
         <appender-ref ref="STDOUT"/>
     </logger>
 
-    <logger name="akka.projection.r2dbc" level="DEBUG" />
-    <logger name="akka.persistence.r2dbc" level="DEBUG" />
+    <logger name="akka.projection.r2dbc" level="TRACE" />
+    <logger name="akka.persistence.r2dbc" level="TRACE" />
```
johanandren, line 16: Do we want to leave at trace?

johanandren (commented):
Missing Instant JDK API in pre 17 or something it seems.

projection/src/test/resources/logback-test.xml
patriknw: thanks, reverted to debug

ennru (approved):
LGTM.

johanandren (approved):

PR #325, opened 2022-11-28
ci: check out full history for Fossa analyze

ennru:
Fossa runs `sbt makePom` internally and needs more git history to make MiMa find the latest version.


patriknw (approved):
LGTM

PR #326, opened 2022-11-30
feat: Cleanup tool

patriknw:
* to support deletes of events and snapshots

TODO:
- [x] javadsl
- [x] DurableStateCleanup

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/cleanup/javadsl/EventSourcedCleanup.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/cleanup/javadsl/EventSourcedCleanup.scala
@@ -1,0 +1,102 @@
+/*
+ * Copyright (C) 2022 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.r2dbc.cleanup.javadsl
+
+import java.util.concurrent.CompletionStage
+
+import scala.collection.immutable
+import scala.compat.java8.FutureConverters._
+
+import akka.Done
+import akka.actor.ClassicActorSystemProvider
+import akka.annotation.ApiMayChange
+import akka.persistence.r2dbc.cleanup.scaladsl
+
+/**
+ * Java API: Tool for deleting all events and/or snapshots for a given list of `persistenceIds` without using persistent
+ * actors. It's important that the actors with corresponding `persistenceId` are not running at the same time as using
+ * the tool.
+ *
+ * WARNING: deleting events is generally discouraged in event sourced systems.
+ *
+ * If `neverUsePersistenceIdAgain` is `true` then the highest used sequence number is deleted and the `persistenceId`
+ * should not be used again, since it would be confusing to reuse the same sequence numbers for new events.
+ *
+ * When a list of `persistenceIds` are given they are deleted sequentially in the order of the list. It's possible to
+ * parallelize the deletes by running several cleanup operations at the same time operating on different sets of
+ * `persistenceIds`.
+ */
+@ApiMayChange
+final class EventSourcedCleanup private (delegate: scaladsl.EventSourcedCleanup) {
+
+  def this(systemProvider: ClassicActorSystemProvider, configPath: String) =
+    this(new scaladsl.EventSourcedCleanup(systemProvider, configPath))
+
+  def this(systemProvider: ClassicActorSystemProvider) =
+    this(systemProvider, "akka.persistence.r2dbc.cleanup")
+
+  /**
+   * Delete all events before a sequenceNr for the given persistence id. Snapshots are not deleted.
+   *
+   * @param persistenceId
+   *   the persistence id to delete for
+   * @param toSequenceNr
+   *   sequence nr (inclusive) to delete up to
+   */
+  def deleteEventsTo(persistenceId: String, toSequenceNr: Long): CompletionStage[Done] =
+    delegate.deleteEventsTo(persistenceId, toSequenceNr).toJava
+
+  /**
+   * Delete all events related to one single `persistenceId`. Snapshots are not deleted.
+   */
+  def deleteAllEvents(persistenceId: String, neverUsePersistenceIdAgain: Boolean): CompletionStage[Done] =
+    delegate.deleteAllEvents(persistenceId, neverUsePersistenceIdAgain).toJava
+
+  /**
+   * Delete all events related to the given list of `persistenceIds`. Snapshots are not deleted.
+   */
+  def deleteAllEvents(
+      persistenceIds: immutable.Seq[String],
+      neverUsePersistenceIdAgain: Boolean): CompletionStage[Done] =
+    delegate.deleteAllEvents(persistenceIds, neverUsePersistenceIdAgain).toJava
+
+  /**
+   * Delete snapshots related to one single `persistenceId`. Events are not deleted.
+   */
+  def deleteSnapshot(persistenceId: String): CompletionStage[Done] =
+    delegate.deleteSnapshot(persistenceId).toJava
+
+  /**
+   * Delete all snapshots related to the given list of `persistenceIds`. Events are not deleted.
+   */
+  def deleteSnapshots(persistenceIds: immutable.Seq[String]): CompletionStage[Done] =
+    delegate.deleteSnapshots(persistenceIds).toJava
+
+  /**
+   * Deletes all events for the given persistence id from before the snapshot. The snapshot is not deleted. The event
+   * with the same sequence number as the remaining snapshot is deleted.
+   */
+  def cleanupBeforeSnapshot(persistenceId: String): CompletionStage[Done] =
+    delegate.cleanupBeforeSnapshot(persistenceId).toJava
+
+  /**
+   * See single persistenceId overload for what is done for each persistence id
+   */
+  def cleanupBeforeSnapshot(persistenceIds: immutable.Seq[String]): CompletionStage[Done] =
+    delegate.cleanupBeforeSnapshot(persistenceIds).toJava
+
+  /**
+   * Delete everything related to one single `persistenceId`. All events and snapshots are deleted.
+   */
+  def deleteAll(persistenceId: String, neverUsePersistenceIdAgain: Boolean): CompletionStage[Done] =
+    delegate.deleteAll(persistenceId, neverUsePersistenceIdAgain).toJava
+
+  /**
+   * Delete everything related to the given list of `persistenceIds`. All events and snapshots are deleted.
+   */
+  def deleteAll(persistenceIds: immutable.Seq[String], neverUsePersistenceIdAgain: Boolean): CompletionStage[Done] =
+    delegate.deleteAll(persistenceIds, neverUsePersistenceIdAgain).toJava
+
+}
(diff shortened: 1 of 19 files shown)


core/src/main/scala/akka/persistence/r2dbc/cleanup/javadsl/EventSourcedCleanup.scala
```diff
@@ -0,0 +1,102 @@
...
+   * Deletes all events for the given persistence id from before the snapshot. The snapshot is not deleted. The event
+   * with the same sequence number as the remaining snapshot is deleted.
+   */
+  def cleanupBeforeSnapshot(persistenceId: String): CompletionStage[Done] =
+    delegate.cleanupBeforeSnapshot(persistenceId).toJava
+
+  /**
+   * See single persistenceId overload for what is done for each persistence id
+   */
+  def cleanupBeforeSnapshot(persistenceIds: immutable.Seq[String]): CompletionStage[Done] =
```
johanandren, line 87: Scala collection in Java DSL, one more below

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -242,46 +242,60 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
+              .bind(9, true)
+          }
+
+          r2dbcExecutor
+            .update(s"delete [$persistenceId]") { connection =>
+              val deleteStatement = connection
+                .createStatement(deleteEventsSql)
+                .bind(0, persistenceId)
+                .bind(1, toSequenceNr)
+              if (neverUsePersistenceIdAgain)
```
johanandren, line 286: Will never happen because of the outermost if-else?

johanandren (commented):
Looks good so far!

core/src/main/scala/akka/persistence/r2dbc/cleanup/javadsl/EventSourcedCleanup.scala
patriknw: thanks, fixed

patriknw:
ready for final review

johanandren (approved):
LGTM

docs/src/main/paradox/cleanup.md
```diff
@@ -0,0 +1,55 @@
...
+
+If possible, it is best to keep all events in an event sourced system. That way new @ref[Projections](projection.md)
+can be re-built.
+
+In some cases keeping all events is not possible or must be removed for regulatory reasons, such as compliance with
+GDPR. `EventSourcedBehavior`s can automatically snapshot state and delete events as described in the
+@extref:[Akka docs](akka:typed/persistence-snapshot.html#snapshot-deletion). Snapshotting is useful even if events
+aren't deleted as it speeds up recovery.
+
+Deleting all events immediately when an entity has reached it's terminal deleted state would have the consequence that
```
pvlugter, line 20:
```suggestion
Deleting all events immediately when an entity has reached its terminal deleted state would have the consequence that
```

docs/src/main/paradox/cleanup.md
```diff
@@ -0,0 +1,55 @@
...
+Projections might not have consumed all previous events yet and will not be notified of the deleted event. Instead, it's
+recommended to emit a final deleted event and store the fact that the entity is deleted separately via a Projection.
+Then a background task can cleanup the events and snapshots for the deleted entities by using the
```
pvlugter, line 23: `cleanup` noun -> `clean up` verb

```suggestion
Then a background task can clean up the events and snapshots for the deleted entities by using the
```

docs/src/main/paradox/cleanup.md
```diff
@@ -0,0 +1,55 @@
+# Database Cleanup
+
+## Event Sourced cleanup tool
+
+@@@ warning
+
+When running any operation with the `EventSourcedCleanup`  for a persistence id the actor with that persistence id must
```
pvlugter, line 7: Maybe either use "with the `EventSourcedCleanup` tool" or just "with `EventSourcedCleanup`"

```suggestion
When running any operation with `EventSourcedCleanup` for a persistence id, the actor with that persistence id must
```

docs/src/main/paradox/cleanup.md
```diff
@@ -0,0 +1,55 @@
...
+: @@snip [cleanup](/docs/src/test/java/jdocs/home/cleanup/CleanupDocExample.java) { #cleanup }
+
+Scala
+: @@snip [cleanup](/docs/src/test/scala/docs/home/cleanup/CleanupDocExample.scala) { #cleanup }
+
+## Durable State cleanup tool
+
+@@@ warning
+
+When running any operation with the `DurableStateCleanup`  for a persistence id the actor with that persistence id must
```
pvlugter, line 46:
```suggestion
When running any operation with `DurableStateCleanup` for a persistence id, the actor with that persistence id must
```

docs/src/main/paradox/durable-state-store.md
```diff
@@ -1,33 +1,43 @@
...
 
 ## Deletes
 
-The journal supports deletes through hard deletes, which means the durable state store entries are actually deleted from the database.
-There is no materialized view with a copy of the state so make sure to not delete durable states too early if they are used from projections or queries.
+The store supports deletes through hard deletes, which means the durable state store entries are actually deleted from
+the database. There is no materialized view with a copy of the state so make sure to not delete durable states too early
+if they are used from projections or queries.
+
+For each persistent id one tombstone record is kept in the event journal when the state of the persistence id has been
```
pvlugter, line 38:
```suggestion
For each persistent id one tombstone record is kept in the store when the state of the persistence id has been
```

or `durable state store`?

docs/src/main/paradox/journal.md
```diff
@@ -16,18 +18,26 @@ To enable the journal plugin to be used by default, add the following line to yo
...
 
 ## Deletes
 
-The journal supports deletes through hard deletes, which means the journal entries are actually deleted from the database. 
-There is no materialized view with a copy of the event so make sure to not delete events too early if they are used from projections or queries.
+The journal supports deletes through hard deletes, which means the journal entries are actually deleted from the
+database. There is no materialized view with a copy of the event so make sure to not delete events too early if they are
+used from projections or queries.
+
+For each persistent id one tombstone record is kept in the event journal when all events of a persistence id has been
```
pvlugter, line 38:
```suggestion
For each persistent id one tombstone record is kept in the event journal when all events of a persistence id have been
```

docs/src/main/paradox/journal.md
```diff
@@ -16,18 +18,26 @@ To enable the journal plugin to be used by default, add the following line to yo
...
+deleted. The reason for the tombstone record is to keep track of the latest sequence number so that subsequent events
+don't reuse same sequence numbers that have been deleted.
```
pvlugter, line 40:
```suggestion
don't reuse the same sequence numbers that have been deleted.
```

pvlugter (approved):
LGTM

Some small fixes to the docs suggested when reading through.

core/src/main/scala/akka/persistence/r2dbc/cleanup/scaladsl/EventSourcedCleanup.scala
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Scala API: Tool for deleting all events and/or snapshots for a given list of `persistenceIds` without using
+ * persistent actors. It's important that the actors with corresponding `persistenceId` are not running at the same time
+ * as using the tool.
+ *
+ * WARNING: deleting events is generally discouraged in event sourced systems.
+ *
+ * If `neverUsePersistenceIdAgain` is `true` then the highest used sequence number is deleted and the `persistenceId`
+ * should not be used again, since it would be confusing to reuse the same sequence numbers for new events.
```
aludwiko, line 33: `neverUse` might indicate that we will somehow block/lock this `persistenceId`, but this is actually the user's responsibility to not use this persistence id again. Actually, in some cases, it might be perfectly fine to use it again and start seq num from the begging. How about renaming it to `resetSequenceNumber`?

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -242,46 +242,58 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
+              .bind(9, true)
+          }
+
+          r2dbcExecutor
+            .update(s"delete [$persistenceId]") { connection =>
+              Vector(
+                connection
+                  .createStatement(deleteEventsSql)
+                  .bind(0, persistenceId)
+                  .bind(1, toSequenceNr),
```
aludwiko, line 286: Maybe I'm missing sth, but for very long journals for a given persistenceId, deleting millions of rows in one transaction will be very problematic. Big deletes are pretty slow and create a lot of contention for other statements.

core/src/main/scala/akka/persistence/r2dbc/cleanup/scaladsl/EventSourcedCleanup.scala
patriknw: sounds like a good suggestion

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
patriknw: If we would add support for it here we would lookup lowest and highest sequence number and then split that up in multiple transactions executed sequentially in seq number order. Do you have other suggestion? Maybe not that difficult to add? Would you like to take a stab at that in a separate PR?

aludwiko: Yeah, that was my idea as well. To run it in some recursive/iterate way. I could give it a try:)

core/src/main/scala/akka/persistence/r2dbc/cleanup/scaladsl/DurableStateCleanup.scala
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Delete the state related to one single `persistenceId`.
+   */
+  def deleteState(persistenceId: String, neverUsePersistenceIdAgain: Boolean): Future[Done] = {
+    if (neverUsePersistenceIdAgain)
+      stateDao.deleteState(persistenceId, revision = 0L) // hard delete without revision check
+    else {
+      stateDao.readState(persistenceId).flatMap {
+        case None    => Future.successful(Done) // already deleted
+        case Some(s) => stateDao.deleteState(persistenceId, s.revision + 1)
```
aludwiko, line 69: I think in this case we will not delete all previous revisions from the DB. It looks like we need a separate dao delete method with hard delete for both cases.

patriknw: I'm not sure I understand. This is the case where resetSeqNr=false, i.e it should keep a tombstone record which will have the given revision. The value (state) is set to null in the db row. I think this is covered by test "delete events for one persistenceId, but keep seqNr"

(that test case name has a copy-pasta mistake, it should say state not events)

Note that there are no stored previous revisions. Durable State is ordinary in-place updates that only has one row for latest value (revision).

btw, this means that your concern for deleting to many rows in one transaction is only a problem for event sourced, not durable state

aludwiko: "Note that there are no stored previous revisions. Durable State is ordinary in-place updates that only has one row for latest value (revision)." - right! Thanks for the clarification.

Issue #234, opened 2022-05-02
Failed: EventsBySliceSpec


johanandren:
```
[info] - should retrieve from several slices *** FAILED *** (12 seconds, 295 milliseconds)
[1624](https://github.com/akka/akka-persistence-r2dbc/runs/6253686705?check_suite_focus=true#step:7:1624)
[info]   A timeout occurred waiting for a future to complete. Waited 10000000000 nanoseconds. (EventsBySliceSpec.scala:373)
[1625](https://github.com/akka/akka-persistence-r2dbc/runs/6253686705?check_suite_focus=true#step:7:1625)
[info]   org.scalatest.concurrent.ScalaFutures$$anon$1$$anon$2:
[1626](https://github.com/akka/akka-persistence-r2dbc/runs/6253686705?check_suite_focus=true#step:7:1626)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:339)
[1627](https://github.com/akka/akka-persistence-r2dbc/runs/6253686705?check_suite_focus=true#step:7:1627)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[1628](https://github.com/akka/akka-persistence-r2dbc/runs/6253686705?check_suite_focus=true#step:7:1628)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
[1629](https://github.com/akka/akka-persistence-r2dbc/runs/6253686705?check_suite_focus=true#step:7:1629)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:281)
[1630](https://github.com/akka/akka-persistence-r2dbc/runs/6253686705?check_suite_focus=true#step:7:1630)
[info]   at akka.persistence.r2dbc.query.EventsBySliceSpec$$anon$10.<init>(EventsBySliceSpec.scala:373)
[1631](https://github.com/akka/akka-persistence-r2dbc/runs/6253686705?check_suite_focus=true#step:7:1631)
[info]   at akka.persistence.r2dbc.query.EventsBySliceSpec.$anonfun$new$30(EventsBySliceSpec.scala:331)
```
[logs_598.zip](https://github.com/akka/akka-persistence-r2dbc/files/8601588/logs_598.zip)

patriknw:
[...]

PR #329, opened 2022-12-05
feat: cleanup - delete events in batches

aludwiko:


core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -242,58 +249,98 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
+        }
+    }
 
-    result.map(_ => ())(ExecutionContexts.parasitic)
+    if (resetSequenceNumber) {
+      for {
+        fromSeqNr <- readLowestSequenceNr(persistenceId)
+        toSeqNr <- highestSeqNr(persistenceId, toSequenceNr)
+        _ <- deleteInBatches(fromSeqNr, toSeqNr)
+      } yield ()
```
aludwiko, line 335: Is it a problem to use yield, instead of `result.map(_ => ())(ExecutionContexts.parasitic)` ?

patriknw: Might not do that final map with the parasitic but shouldn't matter here. Not that performance critical, and it's the db operations that will take time. yield is fine.

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -242,58 +249,99 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
+            .bind(8, Array.emptyByteArray)
+            .bind(9, true)
+        }
+    }
 
-    result.map(_ => ())(ExecutionContexts.parasitic)
+    if (resetSequenceNumber) {
+      for {
+        fromSeqNr <- readLowestSequenceNr(persistenceId)
+        toSeqNr <- highestSeqNr(persistenceId, toSequenceNr)
```
patriknw, line 334: If we read highest first then we can decide if the readLowestSequenceNr is needed or not. If highest is less than the batch size we know that there is no need for batching and lowest is not needed (or can be set to 1L).

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -242,58 +249,99 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
+        _ <- deleteInBatches(fromSeqNr, toSeqNr)
+      } yield ()
+    } else {
+      for {
+        fromSeqNr <- readLowestSequenceNr(persistenceId)
+        toSeqNr <- highestSeqNr(persistenceId, toSequenceNr)
+        _ <- deleteInBatches(fromSeqNr, toSeqNr)
+        _ <- insertDeleteMarker(toSeqNr)
```
patriknw, line 342: This is  a problem. The delete marker must be in the same transaction as the last delete batch. Otherwise the delete batch might succeed but not the delete marker and we end up with all events deleted but no delete marker.

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -242,58 +249,99 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
-            .bind(1, toSequenceNr)
+            .bind(1, from)
+            .bind(2, to)
         }
-      } else {
+        .map(deletedRows =>
+          if (log.isDebugEnabled) {
+            log.debug(
+              "Deleted [{}] events for persistenceId [{}], from seq num [{}] to [{}]",
+              deletedRows.toString, //toString for fix scala 2.12 compilation error
```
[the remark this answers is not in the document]
aludwiko: done

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
aludwiko: fixed

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
aludwiko: fixed

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -242,58 +249,99 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
-                  .createStatement(deleteEventsSql)
-                  .bind(0, persistenceId)
-                  .bind(1, toSequenceNr),
-                bindDeleteMarker(connection.createStatement(insertDeleteMarkerSql)))
-            }
-            .map(_.head)
-        }
+    val batchSize = journalSettings.cleanupSettings.eventsJournalDeleteBatchSize
+
+    def deleteInBatches(from: Long, maxTo: Long): Future[Unit] = {
```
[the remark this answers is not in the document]
patriknw: Add some tests to EventSourcedCleanupSpec. Configure the test with a smaller batch size than default, but larger than what is used for the existing tests. Write more events and then verify that things are deleted in similar ways as existing tests:
* "delete events for one persistenceId"
* "delete events for one persistenceId, but keep seqNr"

If you also want to verify that it batches as expected I agree that it would be easiest to do that with the `LoggingTestKit`.

patriknw (commented):
Looking good, some test coverage remaining.

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
```diff
@@ -242,58 +251,100 @@ private[r2dbc] class JournalDao(journalSettings: R2dbcSettings, connectionFactor
...
+  }
+
+  private def highestSeqNr(persistenceId: String, toSequenceNr: Long): Future[Long] = {
+    if (toSequenceNr == Long.MaxValue)
+      readHighestSequenceNr(persistenceId, 0L)
+    else
+      Future.successful(toSequenceNr)
+  }
+
+  def lowestSequenceNr(persistenceId: String, toSeqNr: Long, batchSize: Int): Future[Long] = {
```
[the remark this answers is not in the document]
aludwiko: fixed

core/src/main/scala/akka/persistence/r2dbc/journal/JournalDao.scala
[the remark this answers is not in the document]
aludwiko: added, not very beautiful ones, but at least I covered the new functionality

patriknw:
Failed test was unrelated https://github.com/akka/akka-persistence-r2dbc/issues/234
I restarted

patriknw (approved):
LGTM, thanks

PR #330, opened 2022-12-09
perf: only-one-snapshot config

patriknw:
References https://github.com/akka/akka/pull/31784

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -48,6 +48,9 @@ akka.persistence.r2dbc {
 
     # Otherwise it would be a pinned dispatcher, see https://github.com/akka/akka/issues/31058
     plugin-dispatcher = "akka.actor.default-dispatcher"
+
+    # Enables an optimization in Akka for avoiding snapshot deletes in retention.
+    only-one-snapshot = true
   }
 }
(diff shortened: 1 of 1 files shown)


johanandren (approved):

core/src/main/resources/reference.conf
```diff
@@ -48,6 +48,9 @@ akka.persistence.r2dbc {
 
     # Otherwise it would be a pinned dispatcher, see https://github.com/akka/akka/issues/31058
     plugin-dispatcher = "akka.actor.default-dispatcher"
+
+    # Enables an optimization in Akka for avoiding snapshot deletes in retention.
+    only-one-snapshot = true
```
aludwiko, line 53: is not recommended (or not possible) to have more than one snapshot per entity in r2dbc?

patriknw: It's not implemented. It will only store one snapshot, replacing (update) previous. I have not seen a need for historical snapshots and it's adding the burden that applications must cleanup old unused snapshots themselves. History of snapshots probably came from the Cassandra implementation where inserts are (much) faster than updates. Maybe it also was a theoretically cool idea that one can replay the history up to old point in time (not always latest), but that is not used in practise.

PR #339, opened 2022-12-22
feat: Opt out of filtered-not-supported for atLeastOnceFlow log entry

johanandren:


projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -316,14 +316,16 @@ private[projection] object R2dbcProjectionImpl {
...
 
     FlowWithContext[Envelope, ProjectionContext]
       .mapAsync(1) { env =>
         offsetStore
           .isAccepted(env)
           .flatMap { ok =>
             if (ok) {
-              if (skipEnvelope(env)) {
+              if (skipEnvelope(env) && warnAboutFilteredEvents) {
                 log.info("atLeastOnceFlow doesn't support of skipping envelopes. Envelope [{}] still emitted.", env)
```
patriknw, line 329: While you are at it, could you change this so that we are not using EventEnvelope.toString. Including persistenceId and seqNr might be enough.

I'm changing toString in https://github.com/akka/akka/pull/31804 but anyway good to not log the full thing here.

projection/src/test/scala/akka/projection/r2dbc/R2dbcProjectionSpec.scala
```diff
@@ -415,7 +415,7 @@ class R2dbcProjectionSpec
         eventuallyExpectError(sinkProbe).getMessage should startWith(concatHandlerFail4Msg)
       }
       projectedValueShouldBe("e1|e2|e3")
-      offsetShouldBe(3L)
+      offsetShouldBe(3L) // here the failure could
```
patriknw, line 418: incomplete comment?

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
```diff
@@ -316,14 +316,16 @@ private[projection] object R2dbcProjectionImpl {
       offsetStore: R2dbcOffsetStore)(implicit
       ec: ExecutionContext,
       system: ActorSystem[_]): FlowWithContext[Envelope, ProjectionContext, Done, ProjectionContext, _] = {
+    val warnAboutFilteredEvents =
+      system.settings.config.getBoolean("akka.projection.r2dbc.warn-about-filtered-events-in-flow")
```
patriknw, line 320: I'd prefer to place it in R2dbcProjectionSettings and pass in to adaptedHandlerForFlow. I see now that I was lazy and never changed it from `case class`. Fortunately the whole R2dbcProjection is still `@ApiMayChange` so we should cleanup that.

projection/src/test/scala/akka/projection/r2dbc/R2dbcProjectionSpec.scala
johanandren: Focus in wrong window :D

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
johanandren: Tricky because `Envelope` here is a type param with no type bounds at all

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
johanandren: Turned normal class in [e7c2fee](https://github.com/akka/akka-persistence-r2dbc/pull/339/commits/e7c2feec867f13b75b94658c54f1926bb5268b92)

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcProjectionImpl.scala
patriknw: I see, with the new Akka version it will be ok

projection/src/main/scala/akka/projection/r2dbc/R2dbcProjectionSettings.scala
```diff
@@ -34,28 +34,96 @@ object R2dbcProjectionSettings {
...
-    offsetTable: String,
-    timestampOffsetTable: String,
-    managementTable: String,
-    useConnectionFactory: String,
-    timeWindow: JDuration,
-    keepNumberOfEntries: Int,
-    evictInterval: JDuration,
-    deleteInterval: JDuration,
-    logDbCallsExceeding: FiniteDuration) {
+final class R2dbcProjectionSettings private (
```
patriknw, line 45: Thanks!

patriknw (approved):
LGTM

Issue #336, opened 2022-12-20
Document how to configure multiple plugins


patriknw:
```
second-r2dbc-plugin = ${akka.persistence.r2dbc}
second-r2dbc-plugin {
  # here you define the same properties as you would in akka.persistence.r2dbc
}
```

Then the plugin is initialized with plugin id such as `second-r2dbc-plugin.query` or `second-r2dbc-plugin.journal`
Same for the Projection config.

PR #340, opened 2022-12-23
doc: How to configure multiple plugins, #336

patriknw:
This was more involved to explain than I thought, so good to finally have it in docs.

References #336


johanandren (approved):
LGTM

Issue #252, opened 2022-06-15
Configuration of number of pub-sub topics


patriknw:
When enabling `publish-events` it's using 1024 (one per slice) topics per entity type. When using many entity types and active queries that will fill up the ddata message size with errors like:

```
akka.remote.OversizedPayloadException: Discarding oversized payload sent to Some(Actor[akka://sys@host:2552/]): max allowed size 262144 bytes. Message type [ActorSelectionMessage(akka.cluster.ddata.Replicator$Internal$DeltaPropagation)].
```

Would be good with a configuration to group slices into less topics.

PR #341, opened 2023-01-09
feat: Configuration of number of pub-sub topics, #252

patriknw:
References #252

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -34,6 +34,18 @@ akka.persistence.r2dbc {
       throughput-collect-interval = 10 seconds
     }
 
+    # Group the slices for an entity type into this number of topics. Most efficient is to use
+    # the same number as number of projection instances. If configured to less than the number of
+    # of projection instances the overhead is that events will be sent more than once and discarded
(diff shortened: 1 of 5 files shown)


core/src/main/resources/reference.conf
```diff
@@ -34,6 +34,18 @@ akka.persistence.r2dbc {
...
+    # the same number as number of projection instances. If configured to less than the number of
+    # of projection instances the overhead is that events will be sent more than once and discarded
+    # on the destination side. If configured to more than the number of projection instances
+    # the events will only be sent once but there is a risk of exceeding the limits of number
+    # of topics that PubSub can handle (e.g. OversizedPayloadException).
+    # Must be between 1 and 1024 and a whole number divisor of 1024 (number of slices).
+    # This configuration can be changed in a rolling update, but there might be some events
+    # that are not delivered via the pub-sub path and instead delivered later by the queries.
+    # This configuration cannot be defined per journal, but is global for the ActorSystem.
+    publish-events-number-of-topics = 128
```
patriknw, line 47: with hindsight it would have been better to structure all `publish-events` properties in its own config section, but not worth breaking

johanandren (approved):
LGTM

PR #345, opened 2023-01-16
fix: PubSub events shouldn't be too far ahead

patriknw:
* because if consumer stores offset from PubSub event that is further ahead than than the backtracking window a subsequent restart of the consumer will not read the possibly missing events from backtracking

* [x] More test coverage

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
@@ -5,6 +5,7 @@
 package akka.persistence.r2dbc.query.scaladsl
 
 import java.time.Instant
+import java.time.{ Duration => JDuration }
 import scala.collection.immutable
 import scala.collection.mutable
 import scala.concurrent.Future
@@ -178,6 +179,10 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
           }
       dbSource
         .mergePrioritized(pubSubSource, leftPriority = 1, rightPriority = 10)
+        .via(
+          skipPubSubTooFarAhead(
+            settings.querySettings.backtrackingEnabled,
+            JDuration.ofMillis(settings.querySettings.backtrackingWindow.toMillis)))
         .via(deduplicate(settings.querySettings.deduplicateCapacity))
     } else
       dbSource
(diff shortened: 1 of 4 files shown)


core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -218,6 +220,41 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+                env :: Nil
+              } else if (t.timestamp == t.readTimestamp && JDuration
+                  .between(latestBacktracking, t.timestamp)
+                  .compareTo(maxAheadOfBacktracking) > 0) {
+                // drop from pubsub when too far ahead from backtracking
+                // FIXME change to trace? Might be verbose at startup, before backtracking is in sync.
+                log.debug2(
+                  "Dropping pubsub event for persistenceId [{}] seqNr [{}] because too far ahead of backtracking.",
+                  env.persistenceId,
+                  env.sequenceNr)
```
johanandren, line 246: Trace sounds good for the per-element ones.

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -218,6 +220,41 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+      .statefulMapConcat(() => {
+        // track backtracking offset
+        var latestBacktracking = Instant.EPOCH
+        env => {
+          env.offset match {
+            case t: TimestampOffset =>
+              if (env.eventOption.isEmpty) {
+                latestBacktracking = t.timestamp
+                env :: Nil
+              } else if (t.timestamp == t.readTimestamp && JDuration
```
johanandren, line 238: Only envelopes originating from pub-sub will have the same read timestamp as write? (maybe add a comment saying that)

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -218,6 +220,41 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+  @InternalApi private[akka] def rejectPubSubTooFarAhead[Event](
+      maxAheadOfBacktracking: JDuration): Flow[EventEnvelope[Event], EventEnvelope[Event], NotUsed] = {
+    Flow[EventEnvelope[Event]]
+      .statefulMapConcat(() => {
+        // track backtracking offset
+        var latestBacktracking = Instant.EPOCH
+        env => {
+          env.offset match {
+            case t: TimestampOffset =>
+              if (env.eventOption.isEmpty) {
```
johanandren, line 235: for gRPC RES transport, empty event is also events not originating in the remote DC but replicated there, so this may not hold to detect backtracking for that use case?

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
patriknw: Yes, that is the check we have already in `R2dbcOffsetStore`. I agree that is rather vague. Would be good with something more explicit defining that its one of: backtracking, normal, pubsub, ...

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
patriknw: um, then I have to look into that. we might have more places where empty is the way to indicate that it's from backtracking, such as the failing check in the offset store

johanandren: This is what I do on the producing side: https://github.com/akka/akka-projection/pull/757/files#diff-7c1a8b28b12936014dbcd01080bb2d9469c6a3de48b13309cc45f97b71414e76R93

johanandren: So the logic in `R2dbcProjectionImpl#loadEnvelope` will at least anyway load each of those payloads it seems, I had missed that before.

johanandren: I wonder if I should just revert/replace that optimization with a specific payload that is not filtered instead in the gRPC transport.

patriknw:
@johanandren this is on top of https://github.com/akka/akka-persistence-r2dbc/pull/348 so this is the branch you should build from. I also merged the Akka PR to main.

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -223,6 +228,50 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+          var latestBacktracking = Instant.EPOCH
+          env => {
+            env.offset match {
+              case t: TimestampOffset =>
+                if (EnvelopeOrigin.fromBacktracking(env)) {
+                  latestBacktracking = t.timestamp
+                  env :: Nil
+                } else if (EnvelopeOrigin.fromPubSub(env) && latestBacktracking == Instant.EPOCH) {
+                  log.trace2(
+                    "Dropping pubsub event for persistenceId [{}] seqNr [{}] because too no event from backtracking yet.",
```
johanandren, line 252:
```suggestion
                    "Dropping pubsub event for persistenceId [{}] seqNr [{}] because no event from backtracking yet.",
```

johanandren (approved):

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
patriknw: thx, fixed

PR #346, opened 2023-01-16
feat: currentPersistenceIds with entity type filtering

aludwiko:
References https://github.com/lightbend/kalix-proxy/issues/1824

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/DurableStateDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/DurableStateDao.scala
@@ -130,8 +130,14 @@ private[r2dbc] class DurableStateDao(settings: R2dbcSettings, connectionFactory:
   private val allPersistenceIdsSql =
     sql"SELECT persistence_id from $stateTable ORDER BY persistence_id LIMIT ?"
 
+  private val persistenceIdsForEntityTypeSql =
+    sql"SELECT persistence_id from $stateTable WHERE persistence_id LIKE ? ORDER BY persistence_id LIMIT ?"
+
   private val allPersistenceIdsAfterSql =
     sql"SELECT persistence_id from $stateTable WHERE persistence_id > ? ORDER BY persistence_id LIMIT ?"
+
+  private val persistenceIdsForEntityTypeAfterSql =
+    sql"SELECT persistence_id from $stateTable WHERE persistence_id LIKE ? AND persistence_id > ? ORDER BY persistence_id LIMIT ?"
 
   private def stateBySlicesRangeSql(
       maxDbTimestampParam: Boolean,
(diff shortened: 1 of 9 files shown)


johanandren (approved):
LGTM

efgpinto (approved):
LGTM

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/DurableStateDao.scala
```diff
@@ -458,6 +464,30 @@ private[r2dbc] class DurableStateDao(settings: R2dbcSettings, connectionFactory:
...
   }
 
+  def persistenceIds(entityType: String, afterId: Option[String], limit: Long): Source[String, NotUsed] = {
+    val result = r2dbcExecutor.select(s"select persistenceIds by entity type")(
+      connection =>
+        afterId match {
+          case Some(after) =>
+            connection
+              .createStatement(persistenceIdsForEntityTypeAfterSql)
+              .bind(0, entityType + "%")
```
pvlugter, line 474: Should it include the `|` separator, just in case one entity type is a prefix for another?

```scala
entityType + "|%"
```

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/DurableStateDao.scala
```diff
@@ -458,6 +464,30 @@ private[r2dbc] class DurableStateDao(settings: R2dbcSettings, connectionFactory:
...
+          case None =>
+            connection
+              .createStatement(persistenceIdsForEntityTypeSql)
+              .bind(0, entityType + "%")
+              .bind(1, limit)
+        },
+      row => row.get("persistence_id", classOf[String]))
+
+    if (log.isDebugEnabled)
+      result.foreach(rows => log.debug("Read [{}] persistence ids by entity type [{}]", rows.size, entityType))
```
patriknw, line 486: might need `log.debug2(` for compat with Scala 2.12

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/DurableStateDao.scala
aludwiko: I didn't add "|" because of this construction method `def of(entityTypeHint: String, entityId: String, separator: String): PersistenceId =` I could add an additional parameter to `def persistenceIds(` to provide the separator. @patriknw   `only supports the default separator` how do we block using different separators in r2dbc?

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/DurableStateDao.scala
patriknw: or might work fine as is (compiler will tell)

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/DurableStateDao.scala
patriknw: As said, r2dbc only supports the default separator. Supporting custom separator would be a different separate issue (big task with little value). Created issue https://github.com/akka/akka-persistence-r2dbc/issues/347

I don't think we have anything checks that the `persistenceId: String` must contain the default separator, so it will probably work to write the events, but some queries that rely on entityType might not work. I think that is fine for this new query as well since that is already the case in many other places.

octonato: Hmm, so potentially a user could pick another separator when defining their `PersistenceId` and get it all broken? Do we have this documented?

aludwiko: Ok, so it's by the convention. I could write some java/scala doc comments that the `currentPersistenceIds` method will work only with a default separator.

aludwiko: Added.

patriknw (approved):
LGTM

Issue #343, opened 2023-01-13
Projection stuck in mid-event log backtracking


johanandren:
Testing gRPC transport for projections I ended up with (multiple cases of) the following:

On the consuming side:
```
Rejecting unknown sequence number [5] for pid [replicated-shopping-cart|a9b85a87-f170-47e1-8a97-504e26217e69|DCA]. Offset: TimestampOffset(2023-01-12T12:27:27.244543Z,1970-01-01T00:00:00Z,Map(replicated-shopping-cart|a9b85a87-f170-47e1-8a97-504e26217e69|DCA -> 5))

Restarting stream due to failure [1]: java.lang.IllegalStateException: Rejected envelope from backtracking, persistenceId [replicated-shopping-cart|a9b85a87-f170-47e1-8a97-504e26217e69|DCA], seqNr [5], due to unknown sequence number. Please report this issue at https://github.com/akka/akka-persistence-r2dbc
java.lang.IllegalStateException: Rejected envelope from backtracking, persistenceId [replicated-shopping-cart|a9b85a87-f170-47e1-8a97-504e26217e69|DCA], seqNr [5], due to unknown sequence number. Please report this issue at https://github.com/akka/akka-persistence-r2dbc
	at akka.projection.r2dbc.internal.R2dbcOffsetStore.$anonfun$isAccepted$5(R2dbcOffsetStore.scala:727)
	at akka.projection.r2dbc.internal.R2dbcOffsetStore.$anonfun$isAccepted$5$adapted(R2dbcOffsetStore.scala:706)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:467)
	at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:63)
	at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:100)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
	at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:94)
	at akka.dispatch.BatchingExecutor$BlockableBatch.run(BatchingExecutor.scala:100)
	at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:49)
	at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(ForkJoinExecutorConfigurator.scala:48)
```

[...]

PR #348, opened 2023-01-18
fix: Filtered event looks like backtracking event, #343

patriknw:
`R2dbcOffsetStore` is supposed to throw exception if a backtracking even is out of sync, but for events from the tail query and pubsub it should return false so that they are tried again when backtracking event arrives. The problem is for filtered events. We can't see if the filtered event is from backtracking or not. Looks like they are always from backtracking.

Filtered events are represented with `eventOption = None, eventMetadata = Some(NotUsed)`.

We need a way to see if the filtered event is from backtracking. That should still result in a hard error.

References #343

TODO:
- [x] actual fix
- [x] add test of filtered event in the tests of isAccepted

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/EnvelopeOrigin.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/EnvelopeOrigin.scala
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2022 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.r2dbc.internal
+
+import akka.annotation.InternalStableApi
+import akka.persistence.query.UpdatedDurableState
+import akka.persistence.query.typed.EventEnvelope
+
+/**
+ * INTERNAL API
+ */
+@InternalStableApi private[akka] object EnvelopeOrigin {
+  val SourceQuery = ""
+  val SourceBacktracking = "BT"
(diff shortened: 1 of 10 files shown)


patriknw:
This will need the change of the EventEnvelope in Akka https://github.com/akka/akka/pull/31817

core/src/main/scala/akka/persistence/r2dbc/internal/EnvelopeOrigin.scala
```diff
@@ -0,0 +1,42 @@
...
+  val SourcePubSub = "PS"
+
+  def fromBacktracking(env: EventEnvelope[_]): Boolean =
+    env.source == SourceBacktracking
+
+  def fromBacktracking(change: UpdatedDurableState[_]): Boolean =
+    change.value == null
+
+  def fromPubSub(env: EventEnvelope[_]): Boolean =
+    fromPubSub(env.offset)
```
johanandren, line 28: Not `env.source == SourcePubSub` ?

johanandren (approved):

core/src/main/scala/akka/persistence/r2dbc/internal/EnvelopeOrigin.scala
patriknw: right, fixed. nice to get rid of that timestamp check

PR #349, opened 2023-01-19
perf: Avoid too large bursts of backtracking events

patriknw:
https://github.com/akka/akka-persistence-r2dbc/pull/300 added some fairness to switch from backtracking earlier.
That is reverted in https://github.com/akka/akka-persistence-r2dbc/pull/345

This PR solves the original issue in a better way. Instead of only having a time based switch (half backtracking window) it is now also switching to backtracking if many events have been received by the ordinary queries.

### DIFF
--- a/core/src/test/scala/akka/persistence/r2dbc/query/EventsBySlicePerfSpec.scala
+++ b/core/src/test/scala/akka/persistence/r2dbc/query/EventsBySlicePerfSpec.scala
@@ -15,21 +15,43 @@ import akka.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit
 import akka.actor.typed.ActorSystem
 import akka.persistence.query.NoOffset
 import akka.persistence.query.PersistenceQuery
+import akka.persistence.query.TimestampOffset
 import akka.persistence.r2dbc.TestActors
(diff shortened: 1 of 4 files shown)


core/src/test/scala/akka/persistence/r2dbc/query/EventsBySlicePerfSpec.scala
```diff
@@ -83,22 +106,98 @@ class EventsBySlicePerfSpec
...
         val total = Await.result(Future.sequence(counts).map(_.sum), 30.seconds)
         total shouldBe totalNumberOfEvents
         println(
-          s"#$iteration Reading all [$totalNumberOfEvents] events from [${ranges.size}] eventsBySlices " +
+          s"#$iteration Reading all [$totalNumberOfEvents] events from [${ranges.size}] slices with currentEventsBySlices " +
           s"took [${(System.nanoTime() - t1) / 1000 / 1000}] ms")
       }
     }
 
+    "write and read concurrently" in {
```
patriknw, line 129: Used this new test and looked at logs to verify that the queries were executed as expected.

johanandren (approved):
LGTM, and the pile of PRs tested for ~1h intense use without anything weird showing up

Issue #351, opened 2023-01-24
Convenience provider factory for gRPC transport


johanandren:
Now requires a many-param-lambda. A factory, something like `R2dbcProjection.replicationProvider()` would help a lot for DX.
https://github.com/akka/akka-projection/pull/757#discussion_r1084654144

PR #352, opened 2023-01-25
feat: Convenience factories for using with Akka Projection gRPC

johanandren:
References #351

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -35,6 +35,7 @@ object Dependencies {
     val akkaTestkit = "com.typesafe.akka" %% "akka-actor-testkit-typed" % AkkaVersion % Test
     val akkaStreamTestkit = "com.typesafe.akka" %% "akka-stream-testkit" % AkkaVersion % Test
     val akkaJackson = "com.typesafe.akka" %% "akka-serialization-jackson" % AkkaVersion % Test
+    val akkaDiscovery = "com.typesafe.akka" %% "akka-discovery" % AkkaVersion % Test
 
     val akkaProjectionEventSourced =
       "com.lightbend.akka" %% "akka-projection-eventsourced" % AkkaProjectionVersion % Test
@@ -79,6 +80,7 @@ object Dependencies {
     TestDeps.akkaTestkit,
     TestDeps.akkaProjectionTestKit,
     TestDeps.akkaJackson,
+    TestDeps.akkaDiscovery,
     TestDeps.logback,
     TestDeps.scalaTest)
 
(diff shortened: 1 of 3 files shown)


project/Dependencies.scala
```diff
@@ -71,12 +73,14 @@ object Dependencies {
...
     r2dbcPostgres,
     akkaProjectionCore,
+    akkaProjectionGrpc,
     TestDeps.akkaProjectionEventSourced,
     TestDeps.akkaProjectionDurableState,
     TestDeps.akkaStreamTestkit,
     TestDeps.akkaTestkit,
     TestDeps.akkaProjectionTestKit,
     TestDeps.akkaJackson,
+    TestDeps.akkaDiscovery,
```
patriknw, line 83: why is this needed?

patriknw (commented):
nice

project/Dependencies.scala
johanandren: Not sure, tests started failing with version mismatch between discovery and the rest of the Akka deps, but Test/dependencyTree doesn't show it 🤔 

Does Akka gRPC pull it in for client service discovery perhaps? But no-show in dep tree is a mystery to me.

project/Dependencies.scala
```diff
@@ -71,12 +73,14 @@ object Dependencies {
     r2dbcPool,
     r2dbcPostgres,
     akkaProjectionCore,
+    akkaProjectionGrpc,
```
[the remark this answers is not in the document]
johanandren: I did mark it optional up in the value definition, also put the factories in separate classes from R2dbcProjection so it isn't needed on classpath to use the existing projection factories.

project/Dependencies.scala
patriknw: ok, I don't know, maybe something that resolves differently for milestones (I think I have seen such problems before with milestones)

patriknw (approved):
LGTM

project/Dependencies.scala
```diff
@@ -20,6 +20,7 @@ object Dependencies {
     val akkaPersistenceQuery = "com.typesafe.akka" %% "akka-persistence-query" % AkkaVersion
 
     val akkaProjectionCore = "com.lightbend.akka" %% "akka-projection-core" % AkkaProjectionVersion
+    val akkaProjectionGrpc = "com.lightbend.akka" %% "akka-projection-grpc" % AkkaProjectionVersion % Optional
```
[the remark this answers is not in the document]
johanandren: Let's switch to provided

PR #355, opened 2023-02-14
bump: Netty 4.1.89 due to CVE-2022-41915

patriknw:


project/Dependencies.scala
```diff
@@ -25,6 +27,13 @@ object Dependencies {
     val r2dbcSpi = "io.r2dbc" % "r2dbc-spi" % "1.0.0.RELEASE" // ApacheV2
     val r2dbcPool = "io.r2dbc" % "r2dbc-pool" % "1.0.0.RELEASE" // ApacheV2
     val r2dbcPostgres = "org.postgresql" % "r2dbc-postgresql" % "1.0.0.RELEASE" // ApacheV2
+
+    // bump Netty dependency due to CVE-2022-41915
+    // FIXME remove this when r2dbc-postgresql includes a later Netty version
+    val nettyHandlerProxy = "io.netty" % "netty-handler-proxy" % NettyVersion
+    val nettyResolverDns = "io.netty" % "netty-resolver-dns" % NettyVersion
+    val nettyResolverDnsMacos = "io.netty" % "netty-resolver-dns-macos" % NettyVersion classifier "osx-x86_64"
```
patriknw, line 35: the dns-macos classifier doesn't work for me, still 4.1.84.Final according to dependencyTree

additionally, might also need `osx-aarch_64` for M1

I'm not sure it will work to override the transitive dependency with classifier?

project/Dependencies.scala
```diff
@@ -25,6 +27,13 @@ object Dependencies {
     val r2dbcSpi = "io.r2dbc" % "r2dbc-spi" % "1.0.0.RELEASE" // ApacheV2
     val r2dbcPool = "io.r2dbc" % "r2dbc-pool" % "1.0.0.RELEASE" // ApacheV2
     val r2dbcPostgres = "org.postgresql" % "r2dbc-postgresql" % "1.0.0.RELEASE" // ApacheV2
+
+    // bump Netty dependency due to CVE-2022-41915
+    // FIXME remove this when r2dbc-postgresql includes a later Netty version
+    val nettyHandlerProxy = "io.netty" % "netty-handler-proxy" % NettyVersion
+    val nettyResolverDns = "io.netty" % "netty-resolver-dns" % NettyVersion
+    val nettyResolverDnsMacos = "io.netty" % "netty-resolver-dns-native-macos" % NettyVersion classifier "osx-x86_64" classifier "osx-aarch_64"
```
patriknw, line 35: I had problem with this first, but I had not used the right artifact name. Seems like the build is not failing when using classifier. Anyway, this works, and generates pom:
```
<dependency>
            <groupId>io.netty</groupId>
            <artifactId>netty-resolver-dns-native-macos</artifactId>
            <version>4.1.89.Final</version>
            <classifier>osx-aarch_64</classifier>
        </dependency>
        <dependency>
            <groupId>io.netty</groupId>
            <artifactId>netty-resolver-dns-native-macos</artifactId>
            <version>4.1.89.Final</version>
            <classifier>osx-x86_64</classifier>
        </dependency>
```

johanandren: I ran tests on m1 against Postgres, seems to work fine! 👍

johanandren (approved):

patriknw:
Also PR upstream: https://github.com/pgjdbc/r2dbc-postgresql/pull/584

project/Dependencies.scala
patriknw: thanks for verifying that

PR #358, opened 2023-02-20
chore: submodule for tests in migration

sebastian-alfers:
Prep work to be able to release `akka-persistence-r2dbc-migration` with scala3.


johanandren (approved):

build.sbt
```diff
@@ -109,9 +108,18 @@ lazy val migration = (project in file("migration"))
       }
       "-Dlogback.configurationFile=logback-main.xml" :: "-Xms1G" :: "-Xmx1G" :: "-XX:MaxDirectMemorySize=256M" :: akkaProperties
     })
+  .dependsOn(core % "compile->compile")
```
patriknw, line 111: isn't this the same as `.dependsOn(core)`?

patriknw (approved):
LGTM, just a minor sbt question

build.sbt
patriknw: that is different, because it also includes tests via `test->test`

Issue #115, opened 2021-11-04
Change of slice distribution


patriknw:
https://github.com/akka/akka-persistence-r2dbc/pull/107 added the slices to the offset store table, which should make it easy to change the slice distribution for projections, but that should be tried out for real.

johanandren:
@patriknw do we need some more trying out here before closing?

patriknw:
The test I added should cover this ticket.

PR #366, opened 2023-03-09
feat: Keep offset per slice, for change of slice distribution, #115

patriknw:
Ready for first review round, but I would like to add a test that simulates changing the slice ranges, which is the purpose of this.

References #115

### DIFF
--- a/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
+++ b/projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
@@ -24,7 +24,6 @@ import akka.dispatch.ExecutionContexts
 import akka.persistence.Persistence
 import akka.persistence.query.DeletedDurableState
 import akka.persistence.query.DurableStateChange
-import akka.persistence.query.Offset
 import akka.persistence.query.TimestampOffset
 import akka.persistence.query.UpdatedDurableState
 import akka.persistence.query.typed.EventEnvelope
@@ -62,7 +61,7 @@ private[projection] object R2dbcOffsetStore {
       fromPubSub: Boolean)
 
   object State {
-    val empty: State = State(Map.empty, Vector.empty, Instant.EPOCH)
+    val empty: State = State(Map.empty, Vector.empty, Instant.EPOCH, 0)
 
     def apply(records: immutable.IndexedSeq[Record]): State = {
       if (records.isEmpty) empty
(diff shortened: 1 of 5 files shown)


projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -796,18 +817,20 @@ private[projection] class R2dbcOffsetStore(
         val until = currentState.latestTimestamp.minus(settings.timeWindow)
         val minSlice = timestampOffsetBySlicesSourceProvider.minSlice
         val maxSlice = timestampOffsetBySlicesSourceProvider.maxSlice
+        val notInLatestBySlice = currentState.latestBySlice.map { record =>
+          s"${record.slice}-${record.pid}-${record.seqNr}"
+        }.toArray
```
johanandren, line 822: So up to 1024 slice entries depending on what the projection saw?

johanandren (commented):
Looking good

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
patriknw: yes, up to the number of slices in the slice range that the offset store is responsible for (there is one offset store instance per projection instance)

projection/src/main/scala/akka/projection/r2dbc/internal/R2dbcOffsetStore.scala
```diff
@@ -195,8 +208,9 @@ private[projection] class R2dbcOffsetStore(
     VALUES (?,?,?,?,?,?, transaction_timestamp())"""
 
   // delete less than a timestamp
-  private val deleteOldTimestampOffsetSql: String =
-    sql"DELETE FROM $timestampOffsetTable WHERE slice BETWEEN ? AND ? AND projection_name = ? AND timestamp_offset < ?"
+  private val deleteOldTimestampOffsetSql: String = sql"""
+    DELETE FROM $timestampOffsetTable WHERE slice BETWEEN ? AND ? AND projection_name = ? AND timestamp_offset < ?
+    AND NOT (slice || '-' || persistence_id || '-' || seq_nr) = ANY (?)"""
```
patriknw, line 213: Second thought, the slice is probably not needed here since it's anyway string concatenation. I'll take a look at explain analyze.

patriknw:
added more tests, ready for final review

johanandren (approved):

Issue #371, opened 2023-03-13
Missing event in projection


patriknw:
Noticed once in EventSourcedEndToEndSpec when running locally. Attached logs with comments, but can't understand the reason.

[EventSourcedEndToEndSpec.log](https://github.com/akka/akka-persistence-r2dbc/files/10956207/EventSourcedEndToEndSpec.log)

patriknw:
Might have a CI test fail of this in https://github.com/akka/akka-persistence-r2dbc/actions/runs/4403828706/jobs/7712682649#step:7:2132

That is in the "ordinary" scenario, so not stopping/starting changing slice distribution.

patriknw:
I have a reproducer with trace log level and see what is going wrong. This is a bug that may cause events to be missed in some rare conditions.

PR #373, opened 2023-03-13
fix: Save offset for pid/seqNr of the processed Envelope, #371

patriknw:
* instead of using the pid/seqNr from the seen Map in the TimestampOffset
* otherwise it may save offset for "unrelated" pid/seqNr when events have the same timestamp
* it was found by test failure in EventSourcedEndToEndSpec when there was a combination of same timestamp, projection start/stop and rejected event (see issue for attached trace logs)

References #371

It's ready for a first review, but I need to do similar changes for the other saveOffsets methods


patriknw:
I have completed this.

johanandren (approved):

PR #377, opened 2023-03-21
feat: Tags in events by slice query

johanandren:


core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/QueryDao.scala
```diff
@@ -283,7 +289,10 @@ private[r2dbc] class QueryDao(settings: R2dbcSettings, connectionFactory: Connec
...
           serManifest = row.get("event_ser_manifest", classOf[String]),
           writerUuid = "", // not need in this query
-          tags = Set.empty, // tags not fetched in queries (yet)
+          tags = row.get("tags", classOf[Array[String]]) match {
+            case null      => Set.empty[String]
+            case tagsArray => tagsArray.toSet
+          },
           metadata = readMetadata(row)))
 
   def eventsByPersistenceId(
```
patriknw, line 298: There is one more below in eventsByPersistenceId, maybe best to update that at the same time so that we don't forget. We don't use that yet because it's using the classic EventEnvelope, but I think we have to add another currentEventsByPersistenceId that is using the new envelope.

patriknw (commented):
looking good

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/QueryDao.scala
johanandren: Yeah, I started and went back on it when I realised that's old envelope API. I'll add another r2dbc-only events by pid for now.

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -276,28 +276,26 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
-        }
-      }
+      toSequenceNr: Long): Source[ClassicEventEnvelope, NotUsed] =
+    internalCurrentEventsByPersistenceId(persistenceId, fromSequenceNr, toSequenceNr)
       .map(deserializeRow)
-      .mapMaterializedValue(_ => NotUsed)
-  }
+
+  @ApiMayChange
+  def currentEventsByPersistenceIdTyped[Event](
```
patriknw, line 284: This is a good start, but I'm afraid we will need a new query interface in Akka, because we need it in Projections without dependency to r2dbc. See https://github.com/akka/akka-projection/pull/816/files#diff-86d4e7a7b1a3c630c066b698442916f3338ba622da7bea05f82719750f5906c0R238

patriknw (commented):
lgtm

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
patriknw: yeah, I have no better suggestion. The EventEnvelope is typded and even though that is in a different package name we must use unique names for the traits and methods because we must be able to mixin both old and new.

project/Dependencies.scala
```diff
@@ -10,7 +10,7 @@ object Dependencies {
   val Scala3 = "3.1.3"
   val Scala2Versions = Seq(Scala213, Scala212)
   val ScalaVersions = Dependencies.Scala2Versions :+ Dependencies.Scala3
-  val AkkaVersion = System.getProperty("override.akka.version", "2.8.0")
+  val AkkaVersion = System.getProperty("override.akka.version", "2.8.0+4-94382770-SNAPSHOT")
```
patriknw, line 13: need to add the sonatype snapshot repository if we want to merge this

johanandren: Not entirely happy with merging a snapshot dependency to main, so maybe we wait a bit and see if there are more stuff to get in a 2.8.1-M1 of Akka?

patriknw: sure, I agree

johanandren: Bumped in [b9583a0](https://github.com/akka/akka-persistence-r2dbc/pull/377/commits/b9583a0804fb2b7d5e5fe2dc791ff0b6e98bb6d7)

PR #382, opened 2023-03-30
fix: include tags in pubsub envelope

patriknw:


franciscolopezsancho (approved):
From what I understand  those missing tags are now in [place](https://github.com/akka/akka-persistence-r2dbc/pull/206/files)
 LGTM

octonato (approved):

PR #388, opened 2023-04-14
test: Add tests for jsonb as column type for payloads

patriknw:
* fixed how empty byte array for delete markers are stored

Follow up on https://github.com/akka/akka-persistence-r2dbc/pull/239

cc @odd


johanandren (approved):
LGTM

odd (commented):
LGTM

odd:
I now see that it was already merged, better late than never I suppose 8-)

PR #389, opened 2023-04-14
docs: Reference docs for jsonb as column type for payloads

patriknw:
Follow up on #239

cc @odd

### DIFF
--- a/docs/src/test/scala/docs/home/state/BlogPostJsonColumn.scala
+++ b/docs/src/test/scala/docs/home/state/BlogPostJsonColumn.scala
@@ -15,10 +15,12 @@ class BlogPostJsonColumn extends AdditionalColumn[BlogPost.State, Json] {
   override def bind(upsert: AdditionalColumn.Upsert[BlogPost.State]): AdditionalColumn.Binding[Json] =
     upsert.value match {
       case s: BlogPost.DraftState =>
+        // a json library would be used here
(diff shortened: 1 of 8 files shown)


docs/src/test/scala/docs/home/state/BlogPostJsonColumn.scala
```diff
@@ -0,0 +1,29 @@
...
+  override val columnName: String = "query_json"
+
+  override def bind(upsert: AdditionalColumn.Upsert[BlogPost.State]): AdditionalColumn.Binding[Json] =
+    upsert.value match {
+      case s: BlogPost.DraftState =>
+        val jsonString = s"""{"title": "${s.content.title}", "published": false}"""
+        val json = Json.of(jsonString)
+        AdditionalColumn.BindValue(json)
+      case s: BlogPost.PublishedState =>
+        val jsonString = s"""{"title": "${s.content.title}", "published": true}"""
```
johanandren, line 22: Fine like this for a sample but note that it is not quite safe as quotes in the title string would break it (vs some proper JSON serialization solution).

johanandren (approved):
LGTM

docs/src/test/scala/docs/home/state/BlogPostJsonColumn.scala
patriknw: yeah, let me add a comment that a json library would be used

PR #391, opened 2023-04-19
feat: support client certificates in SSL connections

pvlugter:
Allow client certificates for SSL connections to be configured. Just passing more of the connection options.


ennru (approved):
LGTM.

johanandren (approved):

patriknw (commented):
lgtm, I first thought it would be more clear to prefix those with `client-` or have a separate `ssl.client` section, but I see that the factory provider settings anyway don't have that structure so we just follow that.

pvlugter:
Yes, just copying the same naming structure as the options (and in postgres).

PR #394, opened 2023-04-27
bump: Akka 2.8.1

johanandren:


octonato:
some Scala 3 compilation error though

johanandren:
We need to bump Scala 3.2 here as well I think

johanandren:
Surprising results when building locally, `+Test/compile` works for all Scala versions, `++3.2.2; Test/compile` does not.

patriknw:
> +Test/compile works for all Scala versions, ++3.2.2; Test/compile

That is because we have mixed cross scala versions in the projects here. Sbt can't handle that for `++3`, but `+Test/compile` works.

johanandren:
Ah, we talked about that before, seems I am unable to store that in brain.

PR #408, opened 2023-06-09
feat: Allow arbitrary H2 options passed through from config

johanandren:
Useful to try things out for us, but probably good for users as well.


core/src/main/resources/reference.conf
```diff
@@ -305,10 +305,17 @@ akka.persistence.r2dbc {
     # DDL or DML included in init together with plugin table creation
     # http://www.h2database.com/html/features.html#execute_sql_on_connection
     additional-init=""
+
+    # Additional H2 options to use when creating the database.
+    # See https://h2database.com/javadoc/org/h2/engine/DbSettings.html for a list of available options
+    additional-options = { }
```
patriknw, line 311: somewhat strange syntax?
```
additional-options {
}
```

PR #409, opened 2023-06-15
feat: Compaction events as starting point for eventsBySlices

patriknw:
Very early draft to explore the idea...

* When consumer is far behind or starting from zero it would be good to not have to replay all events but start from some kind of snapshot, here called compaction event.
* First it loads the highest seqNr of all compaction events.
* Then it's running the ordinary eventsBySlices with an adjusted start offset. Filtering out events for each persistenceId before the known compaction event seqNr.
* It would be an application concern to emit these specially tagged compaction events from the EventSourcedBehavior.

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
@@ -6,10 +6,12 @@ package akka.persistence.r2dbc.query.scaladsl
 
 import java.time.Instant
 import java.time.{ Duration => JDuration }
+
 import scala.collection.immutable
 import scala.collection.mutable
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
+
 import akka.NotUsed
 import akka.actor.ExtendedActorSystem
 import akka.actor.typed.pubsub.Topic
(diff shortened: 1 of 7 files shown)


core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -197,6 +200,66 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+      minSlice: Int,
+      maxSlice: Int,
+      offset: Offset): Source[EventEnvelope[Event], NotUsed] = {
+    // FIXME DRY with above eventsBySlices
+
+    Source
+      .futureSource {
+        val timestampOffset = toTimestampOffset(offset)
+        queryDao
+          .compactionOffsets(entityType, minSlice, maxSlice, timestampOffset.timestamp)
```
patriknw, line 214: Thinking about compaction event vs entity snapshot once more.

Instead of emitting compaction events from the entity, could we load all entity snapshots here and emit them as events (using same seqNr as corresponding event). Would require a transformation function as parameter here that would convert the snapshot to the event type.

Then starting the eventsBySlices query and filtering out earlier events in same way as with the compactionOffsets Map.

patriknw:
Closing this, superseded by https://github.com/akka/akka-persistence-r2dbc/pull/410

PR #410, opened 2023-06-20
feat: Snapshots as starting point for eventsBySlices

patriknw:
This is an alternative to https://github.com/akka/akka-persistence-r2dbc/pull/409 and I think this approach is better because it doesn't add a new concept that must be thought of when implementing the entity. It's using the existing snapshots but transforming them to events.

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresSnapshotDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresSnapshotDao.scala
@@ -4,26 +4,33 @@
 
 package akka.persistence.r2dbc.internal.postgres
 
+import java.time.Instant
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.duration.FiniteDuration
+
+import akka.NotUsed
 import akka.actor.typed.ActorSystem
+import akka.actor.typed.scaladsl.LoggerOps
 import akka.annotation.InternalApi
 import akka.dispatch.ExecutionContexts
-import akka.persistence.Persistence
 import akka.persistence.SnapshotSelectionCriteria
 import akka.persistence.r2dbc.R2dbcSettings
+import akka.persistence.r2dbc.internal.BySliceQuery.Buckets
+import akka.persistence.r2dbc.internal.BySliceQuery.Buckets.Bucket
+import akka.persistence.r2dbc.internal.InstantFactory
 import akka.persistence.r2dbc.internal.PayloadCodec
 import akka.persistence.r2dbc.internal.PayloadCodec.RichRow
 import akka.persistence.r2dbc.internal.PayloadCodec.RichStatement
 import akka.persistence.r2dbc.internal.R2dbcExecutor
 import akka.persistence.r2dbc.internal.SnapshotDao
 import akka.persistence.r2dbc.internal.Sql.Interpolation
-import akka.persistence.typed.PersistenceId
+import akka.stream.scaladsl.Source
 import io.r2dbc.spi.ConnectionFactory
 import io.r2dbc.spi.Row
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
-
-import scala.concurrent.ExecutionContext
-import scala.concurrent.Future
 
 /**
  * INTERNAL API
@@ -46,17 +53,18 @@ private[r2dbc] class PostgresSnapshotDao(settings: R2dbcSettings, connectionFact
 
   protected val snapshotTable = settings.snapshotsTableWithSchema
   private implicit val snapshotPayloadCodec: PayloadCodec = settings.snapshotPayloadCodec
-  private val persistenceExt = Persistence(system)
   private val r2dbcExecutor = new R2dbcExecutor(connectionFactory, log, settings.logDbCallsExceeding)(ec, system)
+  private val queryDao = settings.connectionFactorySettings.dialect.createQueryDao(settings, connectionFactory)
 
   protected def createUpsertSql: String =
     sql"""
       INSERT INTO $snapshotTable
-      (slice, entity_type, persistence_id, seq_nr, write_timestamp, snapshot, ser_id, ser_manifest, meta_payload, meta_ser_id, meta_ser_manifest)
-      VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+      (slice, entity_type, persistence_id, seq_nr, db_timestamp, write_timestamp, snapshot, ser_id, ser_manifest, meta_payload, meta_ser_id, meta_ser_manifest)
+      VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
       ON CONFLICT (persistence_id)
       DO UPDATE SET
         seq_nr = excluded.seq_nr,
+        db_timestamp = excluded.db_timestamp,
         write_timestamp = excluded.write_timestamp,
         snapshot = excluded.snapshot,
         ser_id = excluded.ser_id,
@@ -85,7 +93,7 @@ private[r2dbc] class PostgresSnapshotDao(settings: R2dbcSettings, connectionFact
       else ""
 
     sql"""
-      SELECT persistence_id, seq_nr, write_timestamp, snapshot, ser_id, ser_manifest, meta_payload, meta_ser_id, meta_ser_manifest
+      SELECT slice, entity_type, persistence_id, seq_nr, db_timestamp, write_timestamp, snapshot, ser_id, ser_manifest, meta_payload, meta_ser_id, meta_ser_manifest
       FROM $snapshotTable
       WHERE persistence_id = ?
       $maxSeqNrCondition $minSeqNrCondition $maxTimestampCondition $minTimestampCondition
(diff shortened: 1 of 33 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresSnapshotDao.scala
```diff
@@ -168,40 +207,42 @@ private[r2dbc] class PostgresSnapshotDao(settings: R2dbcSettings, connectionFact
   }
 
   def store(serializedRow: SerializedSnapshotRow): Future[Unit] = {
-    val entityType = PersistenceId.extractEntityType(serializedRow.persistenceId)
-    val slice = persistenceExt.sliceForPersistenceId(serializedRow.persistenceId)
+    queryDao.timestampOfEvent(serializedRow.persistenceId, serializedRow.seqNr).flatMap { eventTimestamp =>
```
patriknw, line 210: some overhead when storing the snapshot because I want to use the same timestamp (offset) as the event so that Projection OffsetStore will handle the snapshot event equal to the corresponding ordinary event.

patriknw: Could do it when reading instead, but then more overhead for each query.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresSnapshotDao.scala
```diff
@@ -231,4 +272,84 @@ private[r2dbc] class PostgresSnapshotDao(settings: R2dbcSettings, connectionFact
     }
   }.map(_ => ())(ExecutionContexts.parasitic)
 
+  override def currentDbTimestamp(): Future[Instant] = {
```
patriknw, line 275: added a bunch of stuff to the SnapshotDao to be able to use the existing BySlice abstraction

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresSnapshotDao.scala
```diff
@@ -231,4 +272,84 @@ private[r2dbc] class PostgresSnapshotDao(settings: R2dbcSettings, connectionFact
...
+      minSlice: Int,
+      maxSlice: Int,
+      fromTimestamp: Instant,
+      toTimestamp: Option[Instant],
+      behindCurrentTime: FiniteDuration,
+      backtracking: Boolean): Source[SerializedSnapshotRow, NotUsed] = {
+    val result = r2dbcExecutor.select(s"select snapshotsBySlices [$minSlice - $maxSlice]")(
+      connection => {
+        val stmt = connection
+          .createStatement(snapshotsBySlicesRangeSql(minSlice, maxSlice))
```
patriknw, line 297: this is the way snapshots are loaded

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -197,6 +231,71 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+   * Same as `eventsBySlices` but with the purpose to use snapshots as starting points and thereby reducing number of
+   * events that have to be loaded. This can be useful if the consumer start from zero without any previously processed
+   * offset or if it has been disconnected for a long while and its offset is far behind.
+   *
+   * First it loads all snapshots with timestamps greater than or equal to the offset timestamp. There is at most one
+   * snapshot per persistenceId. The snapshots are transformed to events with the given `transformSnapshot` function.
+   *
+   * After emitting the snapshot events the ordinary events with sequence numbers after the snapshots are emitted.
+   */
+  def eventsBySlicesStartingFromSnapshots[Snapshot, Event](
```
patriknw, line 244: so we would have to add a new read journal trait to Akka

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -197,6 +231,71 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+        .currentBySlices("snapshotsBySlices", entityType, minSlice, maxSlice, offset)
+
+    val dbSource =
+      bySlice[Event].liveBySlices(
+        "eventsBySlices",
+        entityType,
+        minSlice,
+        maxSlice,
+        offset, // FIXME adjust based on snapshots
+        snapshotOffsets = Map.empty // FIXME collect offsets corresponding to snapshots
```
patriknw, line 263: I want to collect these offsets from the emitted snapshot events from snapshotSource before starting the dbSource. Can that be done with existing Akka Streams operators or do I have to implement a graph stage?

ddl-scripts/create_tables_postgres.sql
```diff
@@ -29,6 +29,7 @@ CREATE TABLE IF NOT EXISTS snapshot(
   entity_type VARCHAR(255) NOT NULL,
   persistence_id VARCHAR(255) NOT NULL,
   seq_nr BIGINT NOT NULL,
+  db_timestamp timestamp with time zone NOT NULL,
```
patriknw, line 32: wouldn't be difficult to alter table for migration

johanandren (commented):
Looks relatively reasonable ;) 

I don't have a better idea.

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
patriknw: thanks, I'll try

ddl-scripts/create_tables_postgres.sql
patriknw: we can remove the NOT NULL here, the fallback can be to use the write_timestamp when it's null. It's probably good enough for the migration scenario. Alternatively lazy loading the timestamp from the event_journal if that column is null. That latter might be better if we are also making the config opt-in on the write side.

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
johanandren: I guess if we anyway side effect it out, doing that directly from `onComplete` in `statefulMap` and then returning `None` is just as good and less messy.

patriknw: The problem is that we return a Source and we are not materializing it ourselves, so it may be materialized more than once and then I can't side effect via a single Promise.

johanandren: You can stuff it all inside a `Source.setup` to tie it to materialization

patriknw: implemented in https://github.com/akka/akka-persistence-r2dbc/pull/410/commits/563a97cfd9a5b7d7a63b4692c8492fbaa176f282

patriknw: ah, I missed the existence of Source.setup. Don't you think it's anyway more clean with a custom graph stage. It wasn't too involved.

johanandren: Yep, looks custom stage like a good path to me. 👍

patriknw:
This is ready for final review

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -168,35 +200,208 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+   *
+   * First it loads all snapshots with timestamps greater than or equal to the offset timestamp. There is at most one
+   * snapshot per persistenceId. The snapshots are transformed to events with the given `transformSnapshot` function.
+   *
+   * After emitting the snapshot events the ordinary events with sequence numbers after the snapshots are emitted.
+   *
+   * To use `eventsBySlicesStartingFromSnapshots` you must enable configuration
+   * `akka.persistence.r2dbc.query.start-from-snapshot.enabled` and follow instructions in migration guide
+   * https://doc.akka.io/docs/akka-persistence-r2dbc/current/migration-guide.html#eventsBySlicesStartingFromSnapshots
+   */
```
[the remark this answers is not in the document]
patriknw: The Query traits are very vague how the implementation should behave, by design. Therefore I think it's important to have more precise api docs in the impl.

For example, this is not something that we can document in Akka:
```
The supported offset is [[akka.persistence.query.TimestampOffset]] and [[Offset.noOffset]].
```

migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
```diff
@@ -283,6 +283,7 @@ class MigrationTool(system: ActorSystem[_]) {
       case Some(selectedSnapshot @ SelectedSnapshot(snapshotMetadata, _)) =>
         for {
           seqNr <- {
+            // TODO we could load the timestamp and tags from corresponding event, see R2dbcSnapshotStore.saveAsync
```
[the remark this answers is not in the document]
patriknw: In MigrationTool I don't think this is important, and that's why I used TODO instead of FIXME. Instead, when enabling the feature they should run the sql migration step as documented.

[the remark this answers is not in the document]
patriknw: I'll change the comment, and mention what I said instead. Not TODO.

johanandren (approved):

patriknw:
Waiting for Akka release that includes https://github.com/akka/akka/pull/31980

johanandren (approved):
LGTM

PR #416, opened 2023-07-03
feat: Turn filtered payload into filtered envelopes

johanandren:


core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -112,7 +118,7 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
         metadata,
         row.entityType,
         row.slice,
-        filtered = false,
+        filtered = event.isEmpty,
```
johanandren, line 121: This is probably not enough/correct because missing payload could also mean lazy loaded for backtracking

johanandren: Added a flag for lazy load in [a359c39](https://github.com/akka/akka-persistence-r2dbc/pull/416/commits/a359c390a6ccb56785c8a4258d3220cee158f857)

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -112,7 +118,7 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
         metadata,
         row.entityType,
         row.slice,
-        filtered = false,
+        filtered = event.isEmpty && !row.payloadLazyLoaded,
```
patriknw: I don't think this is as intended, because we want to preserve the filtered filtered flag also for backtracking events. Fixed in 66135c2

patriknw (commented):
I have pushed a suggestion to avoid addional flag 66135c2

johanandren:
LGTM, but can't do actual review because I opened the PR. Will merge.

PR #419, opened 2023-07-05
fix: Close connection if execution takes too long

patriknw:
* The current transaction will be rolled back and connection closed if exceeded.
* This timeout is needed to handle some failure scenarios, when the database server is terminated in a non graceful way and there is a load balancer in front. The client connection and current execution in progress would not be completed without this timeout, resulting in a "connection leak".

I have tried this by simulating execution that takes longer than the timeout with `pg_sleep`, but I haven't included tests yet. Will make a new test for this.

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -253,6 +253,25 @@ akka.persistence.r2dbc {
     # Enabling this has some performance overhead.
     # A fast query for Postgres is "SELECT 1"
     validation-query = ""
+
+
+    # Abort any statement that takes more than the specified amount of time.
+    # This timeout is handled by the database server.
+    # This timeout should be less than `close-calls-exceeding`.
+    statement-timeout = off
+
+    # Maximum SQL statement execution duration. The current connection is closed if exceeded,
+    # and will not be reused by the pool.
+    # This timeout is handled on the client side and should be used in case the database server
+    # is unresponsive or the connection is broken but not closed.
+    # It can be used in combination with `statement-timeout`, which should be less than this
+    # timeout.
+    # The timeout is needed to handle some failure scenarios, when the database server is
+    # terminated in a non graceful way and there is a load balancer in front. The client
+    # connection and current execution in progress would not be completed without this timeout,
+    # resulting in a "connection leak".
+    # Set to "off" to disable this timeout.
+    close-calls-exceeding = 20 seconds
   }
 
   # If database timestamp is guaranteed to not move backwards for two subsequent
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
@@ -5,13 +5,19 @@
 package akka.persistence.r2dbc.internal
 
 import java.util.function.BiConsumer
+
 import scala.collection.immutable
 import scala.collection.mutable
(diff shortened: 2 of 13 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -301,26 +323,82 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger, logDb
...
+      .andThen { case _ =>
+        try connection.close().asFutureDone()
+        catch ignoreConnectionClosedException
+
+      } catch ignoreConnectionClosedException
+  }
+
+  private def closeAfterTimeout(connection: Connection): Unit = {
+    import scala.concurrent.duration._
+    val timeout = 5.seconds
```
pvlugter, line 365: Make this extra timeout configurable as well, or based on execution timeout?

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -301,26 +323,82 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger, logDb
...
+
+    log.warn("Timeout after [{}], will try to close connection", timeout)
```
pvlugter, line 367:
```suggestion
    log.warn("Timeout after [{}], will try to close connection", executionTimeout)
```

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -301,26 +323,82 @@ class R2dbcExecutor(val connectionFactory: ConnectionFactory, log: Logger, logDb
...
+    // that also times out we will as a last resort try to close the underlying connection
+
+    val closeWrappedTask = connection match {
+      case wrapped: Wrapped[Connection] @unchecked =>
+        Some(
+          system.scheduler.scheduleOnce(
+            timeout,
+            { () =>
+              log.info("Timeout of closing connection (including rollback), will try to close underlying connection")
+              wrapped.close.asFutureDone().onComplete {
```
pvlugter, line 379: Unwrap to get underlying connection?

```suggestion
              wrapped.unwrap.close.asFutureDone().onComplete {
```

pvlugter (commented):
Trying this out with a forced failover, it only partially recovers the connections. Haven't debugged, but could be depending on the state or transaction of the processing statement. And connections may be released into the pool in the hanging state still.

While directly closing the underlying postgres connection looks like it does work cleanly, with connections invalidated.

core/src/main/resources/reference.conf
```diff
@@ -253,6 +253,25 @@ akka.persistence.r2dbc {
     # Enabling this has some performance overhead.
     # A fast query for Postgres is "SELECT 1"
     validation-query = ""
+
+
+    # Abort any statement that takes more than the specified amount of time.
+    # This timeout is handled by the database server.
+    # This timeout should be less than `close-calls-exceeding`.
+    statement-timeout = off
```
patriknw, line 261: I didn't enable this by default, but I would guess that it's good to have, and should probably be slightly less then the circuit breaker timeout (default 10 seconds)

pvlugter:
Testing forced failover, with default settings here, all looks good.

pvlugter (approved):
LGTM

PR #425, opened 2023-07-07
fix: Close connection if execution takes too long (fw port)

patriknw:
* The underlying connection is closed if timeout exceeded, and connection will not be reused by the pool.
* This timeout is needed to handle some failure scenarios, when the database server is terminated in a non graceful way and there is a load balancer in front. The client connection and current execution in progress would not be completed without this timeout, resulting in a "connection leak".
* Also adding config for server side statement timeout
* forward port, cherry picked from commit 6e8b236583fc541d8e89fa53cb23a6291ebaeb9a

Forward port of https://github.com/akka/akka-persistence-r2dbc/pull/419

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/session/scaladsl/R2dbcSession.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/session/scaladsl/R2dbcSession.scala
@@ -8,9 +8,11 @@ import scala.collection.immutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration._
+
 import akka.actor.typed.ActorSystem
 import akka.annotation.ApiMayChange
-import akka.persistence.r2dbc.{ ConnectionFactoryProvider, R2dbcSettings }
+import akka.persistence.r2dbc.ConnectionFactoryProvider
+import akka.persistence.r2dbc.ConnectionPoolSettings
 import akka.persistence.r2dbc.internal.R2dbcExecutor
 import io.r2dbc.spi.Connection
 import io.r2dbc.spi.Row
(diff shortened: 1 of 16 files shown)


core/src/main/scala/akka/persistence/r2dbc/session/scaladsl/R2dbcSession.scala
```diff
@@ -33,9 +35,17 @@ object R2dbcSession {
...
-    val connectionFactory =
-      ConnectionFactoryProvider(system).connectionFactoryFor(connectionFactoryConfigPath)
-    val r2dbcExecutor = new R2dbcExecutor(connectionFactory, log, logDbCallsDisabled)(system.executionContext, system)
+    val connectionFactoryProvider = ConnectionFactoryProvider(system)
+    val connectionFactory = connectionFactoryProvider.connectionFactoryFor(connectionFactoryConfigPath)
+    val closeCallsExceeding =
+      connectionFactoryProvider
+        .connectionFactorySettingsFor(connectionFactoryConfigPath)
+        .poolSettings
+        .closeCallsExceeding
```
patriknw, line 44: I was pondering where to place the closeCallsExceeding setting. It has to be in the connection-factory section in reference.conf because it's used from this `withSession`, which isn't aware of pluginId or such.

It's not specific to a dialect, so therefore I placed it in the common `default-connection-pool`. However, it's not really related to the pool. Maybe it's better to place it directly in `connection-factory` and `ConnectionFactorySettings`?

johanandren: I don't think pool is that wrong, isn't it about keeping the pool healthy?

core/src/test/scala/akka/persistence/r2dbc/internal/R2dbcExecutorSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+
+  // need pg_sleep or similar
+  private def canBeTestedWithDialect: Boolean =
+    r2dbcSettings.connectionFactorySettings.dialect == PostgresDialect ||
+    r2dbcSettings.connectionFactorySettings.dialect == YugabyteDialect
+
+  private def pendingIfCannotBeTestedWithDialect(): Unit = {
+    if (!canBeTestedWithDialect) {
+      info(s"Can't be tested with dialect [${r2dbcSettings.dialectName}]")
+      pending
```
johanandren, line 58: Annoying that there is no `ignore`, for things that aren't really pending to be fixed. (There is `cancel()` but that prints a stacktrace, which is worse than pending)

johanandren (approved):

core/src/main/scala/akka/persistence/r2dbc/session/scaladsl/R2dbcSession.scala
patriknw: yeah, good enough, not exposed to users, the config property would be the same

PR #401, opened 2023-05-10
chore: docker-compose postgres healthcheck

patriknw:
* instead of sleep in CI

Thanks @pvlugter

PR #437, opened 2023-08-07
chore: add docker-compose healthcheck for postgres

pvlugter:
Fresh attempt for #401.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -67,10 +67,8 @@ jobs:
 
       - name: Start DB
         run: |-
-          docker-compose -f docker/docker-compose-postgres.yml up -d
-          # TODO: could we poll the port instead of sleep?
-          sleep 10
-          docker exec -i docker_postgres-db_1 psql -U postgres -t < ddl-scripts/create_tables_postgres.sql
+          docker compose -f docker/docker-compose-postgres.yml up --wait
+          docker exec -i postgres-db psql -U postgres -t < ddl-scripts/create_tables_postgres.sql
 
       - name: sbt test
         run: |-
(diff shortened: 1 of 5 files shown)


docker/docker-compose-postgres.yml
```diff
@@ -2,8 +2,15 @@ version: '2.2'
...
     image: postgres:latest
+    container_name: postgres-db
     ports:
       - 5432:5432
     environment:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
+    healthcheck:
+      test: ['CMD', 'pg_isready', "-q", "-d", "postgres", "-U", "postgres"]
+      interval: 5s
```
patriknw, line 13: doesn't matter much but maybe this should be shorter, like 2s (and more retries)?

.github/workflows/build-test.yml
```diff
@@ -67,10 +67,8 @@ jobs:
 
       - name: Start DB
         run: |-
-          docker-compose -f docker/docker-compose-postgres.yml up -d
-          # TODO: could we poll the port instead of sleep?
-          sleep 10
-          docker exec -i docker_postgres-db_1 psql -U postgres -t < ddl-scripts/create_tables_postgres.sql
+          docker compose -f docker/docker-compose-postgres.yml up --wait
```
patriknw, line 70: and just to repeat, the fix is:

> Looks like there are two versions of docker-compose in the [image](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md), both v1 and v2. Where docker-compose is v1, while using docker compose will use the v2 plugin.

patriknw (approved):
LGTM, thanks for picking up this

docker/docker-compose-postgres.yml
pvlugter: Yes, could be shorter. Just left it the same as earlier.

pvlugter: Looking at this more closely, the healthcheck will keep being tested at this interval while the container is running. So when originally setting this up (for Kalix) the 5s interval was a balance, for getting to healthy but also not checking too frequently during running. Shorter is probably fine though.

There's a new start-interval setting coming, so that the regular healthcheck interval can be longer, and the interval during startup can be much shorter. Maybe we use that, once it's released and available. Scheduled for docker engine version 25.0.

PR #440, opened 2023-08-22
fix: Filter FilteredPayload in pub-sub

johanandren:


patriknw (approved):
LGTM, right it can be published from the EventWriter like that

PR #442, opened 2023-08-24
chore: ConnectionPoolSettings from ConnectionFactoryProvider

patriknw:
* for the new closeCallsExceeding paramater in R2dbcExecutor, needed from R2dbcProjection


core/src/main/scala/akka/persistence/r2dbc/ConnectionFactoryProvider.scala
```diff
@@ -71,6 +72,13 @@ class ConnectionFactoryProvider(system: ActorSystem[_]) extends Extension {
     }
   }
 
+  /**
+   * INTERNAL API
+   */
+  @InternalStableApi
+  def connectionPoolSettingsFor(configLocation: String): ConnectionPoolSettings =
```
patriknw, line 79: ConnectionPoolSettings is already public but with InternalStableApi marker

johanandren (approved):

Issue #444, opened 2023-08-31
Dynamically increasing backtracking window


patriknw:
It can be difficult to configure the optimal `akka.persistence.r2dbc.query.backtracking.window`. If too large there is overhead when starting up since all envelopes (without event payload) must be loaded. If too small there is a risk of hitting "Rejecting unexpected sequence number" error, especially when combined with scaling number of projection instances.

Note that this must also play well with `akka.projection.r2dbc.offset-store.time-window`.

One idea is to be able to increase the backtracking window in case of these errors, so that next restart look further back.

PR #445, opened 2023-09-05
fix: backtracking queries and filtered events

pvlugter:
Backtracking queries can stop too early. This can create a gap between the normal db queries and the backtracking queries that's longer than the backtracking window, and if a projection instance is then restarted, there are events that are completely missed, or rejected because of earlier missed events.

The switch-from-backtracking check compares how many rows were returned by the past query to the buffer size - 1. The 1 is for the event at the initial timestamp, which has already been seen by the previous query and will be filtered from the result. If there are events that share timestamps (because they were emitted together), then there can be multiple events filtered out of the result set. In this case, it will switch from backtracking mode before catching up, and while it can still progress one query at a time, it can fall too far behind.

Fix this by keeping track of the seen count for the latest timestamp when backtracking, and then copying this over as the number of events expected to be filtered for the next query. Added this to the internal state here for now, but could be nicer to have it in TimestampOffset itself — for something compatible, there could be an additional map of seen counts, or first seen sequence numbers, along with the latest seen map.

This should fix the underlying issue for #444.

Based on `release-1.1` first, from testing with Kalix, and can then forward-port to main.

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
@@ -305,8 +305,6 @@ import org.slf4j.Logger
           throw new IllegalArgumentException(
             s"Unexpected offset [$offset] before latestBacktracking [${state.latestBacktracking}].")
 
-        // TODO: Could be nicer to track the seen count in TimestampOffset itself, along with the latest seen map.
-        //       Either as a seen count directly, or as a range of seen sequence numbers for each persistence id.
         val newSeenCount =
           if (offset.timestamp == state.latestBacktracking.timestamp) state.latestBacktrackingSeenCount + 1 else 1
 
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
@@ -31,7 +31,7 @@ import org.slf4j.Logger
 
   object QueryState {
     val empty: QueryState =
-      QueryState(TimestampOffset.Zero, 0, 0, 0, 0, backtrackingCount = 0, TimestampOffset.Zero, Buckets.empty)
+      QueryState(TimestampOffset.Zero, 0, 0, 0, 0, backtrackingCount = 0, TimestampOffset.Zero, 0, 0, Buckets.empty)
   }
 
   final case class QueryState(
@@ -42,6 +42,8 @@ import org.slf4j.Logger
       idleCount: Long,
       backtrackingCount: Int,
       latestBacktracking: TimestampOffset,
+      latestBacktrackingSeenCount: Int,
+      backtrackingExpectFiltered: Int,
       buckets: Buckets) {
 
     def backtracking: Boolean = backtrackingCount > 0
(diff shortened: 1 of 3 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -303,11 +305,30 @@ import org.slf4j.Logger
...
 
+        if (log.isDebugEnabled()) {
+          if (state.latestBacktracking.seen.nonEmpty &&
+            offset.timestamp.isAfter(state.latestBacktracking.timestamp.plus(firstBacktrackingQueryWindow)))
+            log.debugN(
+              "{} next offset is outside the backtracking window, latestBacktracking: [{}], offset: [{}]",
+              logPrefix,
+              state.latestBacktracking,
+              offset)
+        }
```
pvlugter, line 330: Debugging for when the normal query is getting too far ahead. Left it in as could be useful.

Only debugging, as this can still happen if there are no events for over the backtracking window (2 minutes), then when new events arrive they'll be processed by the normal query but won't be seen by backtracking until after the longer backtracking behind-current-time has passed. It's expected behaviour and not a problem in this case — the only unprocessed events are within behind-current-time (10 seconds) and there won't be missed events on restart.

pvlugter (commented):
And assuming it's ok to change the internal QueryState, without making this compatible? Added the mima filters.

pvlugter:
For some reason, the PR has `Run tests with Postgres (2.12, 1.8.0, temurin:1.8.0, test)` as a required check, but that's a nightly test. I'll update the required checks for the `release-*` branches, removing this, and adding all the current tests as required (like for main).

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -303,11 +305,30 @@ import org.slf4j.Logger
           throw new IllegalArgumentException(
             s"Unexpected offset [$offset] before latestBacktracking [${state.latestBacktracking}].")
 
-        state.copy(latestBacktracking = offset, rowCount = state.rowCount + 1)
+        // TODO: Could be nicer to track the seen count in TimestampOffset itself, along with the latest seen map.
+        //       Either as a seen count directly, or as a range of seen sequence numbers for each persistence id.
```
patriknw, line 309: I'm not convinced that would be better, since TimestampOffset is in Akka core and this is rather deep into implementation details of this query.

patriknw (approved):
LGTM, great that you tracked it down 😄

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
pvlugter: Yeah, very specifically used here for the backtracking. I was thinking if it was added to TimestampOffset, then it could be more general, in case it's useful elsewhere for similar reasons. There's the `seen` map already, with the latest sequence number. Could be an additional `firstSeen` map, so that it stays compatible, and either the seen counts, or the range of sequence numbers, can be derived from the two maps. Otherwise, if not applicable elsewhere, can just remove the comment too.

patriknw: Let's keep it in the internal state for now (remove the todo). If we need it elsewhere we can revisit.
It would raise questions like, should the offset store also be able to recreate that firstSeen when the offset is loaded at startup (as far as I see we don't need that).

pvlugter: Right, should not be needed here — for the offset store to load that information. If a loaded offset was being used to initiate a query that was tracking this, then a loaded firstSeen would be useful, as the loaded seen map would be used for filtering. But here the first backtracking offset is created fresh, starting with an empty seen.

johanandren (approved):
LGTM, great fix!

Note that we already have gone into 1.2.0-M-country with main, so to release an 1.1.2 this would need a branch + cherry-pick from somewhere like 58eee5a

patriknw:
we have already branched, this is already targeting `release-1.1`
I can release this today, and forward port to main

PR #453, opened 2023-09-08
H2 successful update not seen by next read

johanandren:
Refs https://github.com/akka/akka-projection/pull/997

Not intended for merging, just pushing for sharing

Trying to emulate what projections does where the missed write is observed. No success to reproduce so far.


johanandren:
One thing I noticed with the two-step update was that H2 leaks memory with the `trace-logging = on` and would OOM the JVM pretty quickly, we should probably add a note in the config around that.

patriknw:
fixed by https://github.com/akka/akka-persistence-r2dbc/pull/456

PR #455, opened 2023-09-14
fix: explicitly set isolation level to read committed

pvlugter:
Refs https://github.com/akka/akka-projection/issues/992.

When trying to reproduce in https://github.com/akka/akka-projection/pull/997, it behaves as if it's not read committed — which is expected to be the [default for H2](https://www.h2database.com/html/advanced.html#transaction_isolation)? But when setting it explicitly, can no longer reproduce the issue. So is it not actually the default?

Experimental draft for further checking and testing.


pvlugter:
No longer needed. I think I've tracked down the race condition underlying this issue. Explicitly setting the isolation level may have altered the timing enough that it was less likely.

PR #456, opened 2023-09-15
fix(h2): disable optimize reuse results to avoid race condition

pvlugter:
Refs https://github.com/akka/akka-projection/issues/992

See https://github.com/akka/akka-projection/issues/992#issuecomment-1720630672 for details.


johanandren (approved):

patriknw (approved):
LGTM 🪨

PR #459, opened 2023-10-02
chore: bump to akka 2.9.0-M2

aludwiko:


patriknw (commented):
any particular reason you bump this? otherwise I'll take care of it as part of the release train. Scala 2.12 and jdk 8 must be dropped first.

patriknw:
Superseded by https://github.com/akka/akka-persistence-r2dbc/pull/462
The reason for compilation error here is that it used a case class copy method of an ApiMayChange class, which we changed.

aludwiko:
I'm experimenting with the latest akka version (2.9.0-M2) and I've just noticed that projections are failing in runtime with akka-persistence-r2dbc 1.2.0-M5.

PR #464, opened 2023-10-03
fix: Envelope source is different for durable state

patriknw:
* NullPointerException in isPayloadDefined
* regression when adding the starting from snapshots, i.e. this bug was not released in any final version (but in milestones)
* for durable state we use null payload as the backtracking marker in the row, becuse we also have the deleted case, ugly but that goes all the way to Akka's UpdatedDurableState

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
@@ -146,7 +146,7 @@ import org.slf4j.Logger
     def seqNr: Long
     def dbTimestamp: Instant
     def readDbTimestamp: Instant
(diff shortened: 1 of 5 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -459,8 +458,7 @@ import org.slf4j.Logger
             behindCurrentTime,
             backtracking = newState.backtracking)
           .filter { row =>
-            val source = if (row.isPayloadDefined) EnvelopeOrigin.SourceQuery else EnvelopeOrigin.SourceBacktracking
```
patriknw, line 462: the NPE was here

johanandren (approved):

PR #481, opened 2023-11-13
chore: change docs project name to disambiguate from  in pom files

ennru:
As it turns out, sbt translates the project name into the pom identifier `artifactId`. This PR changes the project name for the documentation sbt project so that sbt creates unique pom identifiers for the core project and the docs project. This helps other tools to consume the pom files correctly.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -130,7 +130,7 @@ lazy val docs = project
   .settings(common)
   .settings(dontPublish)
   .settings(
-    name := "Akka Persistence R2DBC",
+    name := "Akka Persistence plugin for R2DBC",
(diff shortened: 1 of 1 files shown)


patriknw (approved):
LGTM

build.sbt
```diff
@@ -130,7 +130,7 @@ lazy val docs = project
   .settings(common)
   .settings(dontPublish)
   .settings(
-    name := "Akka Persistence R2DBC",
+    name := "Akka Persistence plugin for R2DBC",
```
patriknw, line 133: for the record, I think the reason we use a readable name here is that it ends up as title or something in docs

PR #483, opened 2023-11-29
chore: remove comment about potential deprecation of the use-app-time config

sebastian-alfers:


core/src/main/resources/reference.conf
```diff
@@ -371,8 +371,6 @@ akka.persistence.r2dbc {
   # Ignored for H2
   db-timestamp-monotonic-increasing = off
 
-  # Enable this for testing or workaround of https://github.com/yugabyte/yugabyte-db/issues/10995
-  # FIXME: This property will be removed when the Yugabyte issue has been resolved.
   # Ignored for H2
```
patriknw, line 374: Add description:
```
# Enable this to generate timestamps from the Akka client side instead of using database timestamps.
# NTP or similar clock synchronization should be used, but some clock skew between clients is accepted.
# Ignored for H2
```

patriknw (approved):
LGTM

PR #485, opened 2023-12-07
feat: Write change event of DurableState to event journal

patriknw:
Early draft to show where the idea is going.

TODO:
- [x] tests
- [x] api in Akka

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
@@ -67,6 +67,7 @@ private[r2dbc] object PostgresDurableStateDao {
       binding: AdditionalColumn.Binding[_])
 
   val FutureDone: Future[Done] = Future.successful(Done)
+  val FutureInstantNone: Future[Option[Instant]] = Future.successful(None)
 }
 
 /**
@@ -302,20 +303,22 @@ private[r2dbc] class PostgresDurableStateDao(
       updatedRows: Long,
       entityType: String,
       change: DurableStateChange[Any],
-      changeEvent: Option[SerializedJournalRow]): Future[Done] = {
+      changeEvent: Option[SerializedJournalRow]): Future[Option[Instant]] = {
     if (updatedRows == 1)
       for {
-        _ <- changeEvent.map(journalDao.writeEventInTx(_, connection)).getOrElse(FutureDone)
+        changeEventTimestamp <- changeEvent
+          .map(journalDao.writeEventInTx(_, connection).map(Some(_)))
+          .getOrElse(FutureInstantNone)
         _ <- changeHandlers.get(entityType).map(processChange(_, connection, change)).getOrElse(FutureDone)
-      } yield Done
+      } yield changeEventTimestamp
     else
-      FutureDone
+      FutureInstantNone
   }
 
   override def upsertState(
       state: SerializedStateRow,
       value: Any,
-      changeEvent: Option[SerializedJournalRow]): Future[Done] = {
+      changeEvent: Option[SerializedJournalRow]): Future[Option[Instant]] = {
     require(state.revision > 0)
 
     def bindTags(stmt: Statement, i: Int): Statement = {
@@ -349,7 +352,7 @@ private[r2dbc] class PostgresDurableStateDao(
 
     val entityType = PersistenceId.extractEntityType(state.persistenceId)
 
-    val result = {
+    val result: Future[(Long, Option[Instant])] = {
       val additionalBindings = additionalColumns.get(entityType) match {
         case None => Vector.empty[EvaluatedAdditionalColumnBindings]
         case Some(columns) =>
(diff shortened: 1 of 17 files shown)


johanandren (commented):
Looks promising

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/R2dbcDurableStateStore.scala
```diff
@@ -125,7 +151,41 @@ class R2dbcDurableStateStore[A](system: ExtendedActorSystem, config: Config, cfg
...
+
+        val entityType = PersistenceId.extractEntityType(persistenceId)
+        val slice = persistenceExt.sliceForPersistenceId(persistenceId)
+        val timestamp = if (journalSettings.useAppTimestamp) InstantFactory.now() else JournalDao.EmptyDbTimestamp
+
+        SerializedJournalRow(
+          slice,
+          entityType,
+          persistenceId,
+          revision,
```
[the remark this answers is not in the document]
johanandren: Can't we just disallow that?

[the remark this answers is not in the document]
patriknw: Do you mean that we would read latest sequence number from journal and thereby check that there is no gap? That has much overhead and adding caching and such is not trivial (as in the EventWriter).

I'm thinking that we might need something to allow starting from later offset without requiring previous for other reasons. E.g. came up in a question about migration from jdbc, or as soon as you have started to delete events and later add a new projection. I'm not sure if it would be some kind of relaxed offset validation or some kind of offset population job that can be run as a migration task.

patriknw:
I think this should be ready for final review now. Will require an Akka release before merging.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
```diff
@@ -64,15 +67,17 @@ private[r2dbc] object PostgresDurableStateDao {
...
  * INTERNAL API
  */
 @InternalApi
-private[r2dbc] class PostgresDurableStateDao(settings: R2dbcSettings, connectionFactory: ConnectionFactory)(implicit
-    ec: ExecutionContext,
-    system: ActorSystem[_])
+private[r2dbc] class PostgresDurableStateDao(
+    settings: R2dbcSettings,
+    connectionFactory: ConnectionFactory,
+    journalDao: JournalDao)(implicit ec: ExecutionContext, system: ActorSystem[_])
```
johanandren, line 80: Would it be worth making the journal-dao creation/dependency more lazy, for use cases where only using "regular" durable state without the change events, this probably means picking up/requiring journal config and maybe even more runtime stuff?

core/src/test/scala/akka/persistence/r2dbc/state/DurableStateUpdateWithChangeEventStoreSpec.scala
```diff
@@ -0,0 +1,188 @@
...
+      val entityType = nextEntityType()
+      val persistenceId = PersistenceId(entityType, "id-to-be-updated-concurrently").id
+      val value = "Genuinely Collaborative"
+      store.upsertObject(persistenceId, revision = 1L, value, tag, s"Changed to $value").futureValue
+
+      val updatedValue = "Open to Feedback"
+      store.upsertObject(persistenceId, revision = 2L, updatedValue, tag, s"Changed to $updatedValue").futureValue
+
+      // simulate an update by a different node that didn't see the first one:
+      val updatedValue2 = "Genuine and Sincere in all Communications"
```
johanandren, line 105: Great values!

johanandren (commented):
Looks good but I wonder about eager dependency on journal.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
patriknw: Two options:
1. reuse the same instance of the JournalDao inside the PostgresDialect
2. pass in the dialect as parmeter here so that it can create the JournalDao on demand

johanandren: Right now we create a separate instance when constructing the durable state dao, can we instead pass a factory lambda and have the journalDao as a lazy field initialized on first use (if that ever happens)?

patriknw: changed in https://github.com/akka/akka-persistence-r2dbc/pull/485/commits/b0b535da3b77461021ce3fb1fc08fa1b7bd6139b

patriknw:
@johanandren I had missed the h2 dialect override and made a mistake of the "in same transaction". Can you take a look at the changes in https://github.com/akka/akka-persistence-r2dbc/pull/485/commits/fbe67e0e349c4d93ff92e2663e1f0f8e5a1cf48a before we merge.

johanandren (approved):

PR #492, opened 2023-12-18
fix: avoid failing when running with testOnly mode

sebastian-alfers:
Before this pr, this test failed when running `sbt 'testOnly *.CurrentPersistenceIdsQuerySpec'` as it causes a false negative for wrongly sorted set cased by single / multi digits edge case for generated ids.

### DIFF
--- a/core/src/test/scala/akka/persistence/r2dbc/query/CurrentPersistenceIdsQuerySpec.scala
+++ b/core/src/test/scala/akka/persistence/r2dbc/query/CurrentPersistenceIdsQuerySpec.scala
@@ -40,7 +40,7 @@ class CurrentPersistenceIdsQuerySpec
 
   private val zeros = "0000"
   private val numberOfEntityTypes = 5
-  private val entityTypes = (1 to numberOfEntityTypes).map(_ => nextEntityType())
+  private val entityTypes = (1 to numberOfEntityTypes).map(n => "Entity" + zeros.drop(n.toString.length) + n)
   private val numberOfPids = 100
   private val pids = {
     (1 to numberOfEntityTypes).flatMap(entityTypeId =>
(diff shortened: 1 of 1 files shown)


core/src/test/scala/akka/persistence/r2dbc/query/CurrentPersistenceIdsQuerySpec.scala
```diff
@@ -65,7 +65,7 @@ class CurrentPersistenceIdsQuerySpec
   "Event Sourced currentPersistenceIds" should {
     "retrieve all ids" in {
       val result = query.currentPersistenceIds().runWith(Sink.seq).futureValue
-      result shouldBe pids.map(_.id)
+      result shouldBe pids.map(_.id).sorted
```
patriknw, line 68: same problem for the other tests in this Spec? I think you can add a sort at the `val pids` instead
```
val pids .... .sortBy(_.id)
```

sebastian-alfers: Yes, I though the same initially (`val pids = ... .sortBy(_.id)`). But when doing that, another spec fails in that file (`retrieve ids for entity type after id`) so I kept it in this single place where all seems to pass fine 🤷

patriknw: alright, the problem is the usage of `nextEntityType`, since for example EntityType-9 is sorted after Entitytype-10, so it depends on where the counter starts at. I think that since this test depends on sorting being the same as in the db it's best to not use that `nextEntityType` here. Instead, create them as:
```
private val entityTypes = (1 to numberOfEntityTypes).map(n => "Entity" + zeros.drop(n.toString.length) + n)
```

Then no further sorting should be needed.

patriknw (approved):
LGTM

johanandren (approved):

PR #505, opened 2024-01-22
feat: sqlserver / mssql support

sebastian-alfers:
Next iteration based on discussion in https://github.com/akka/akka-persistence-r2dbc/pull/494.

---

Note that in the sqlserver specific overrides, named binding is still used. This is recommended by the plugin, and my have several advantages:
- no need to bind a param multiple times if needed,
- no need to use the `getAndIncIndex` approach for conditional binding,
- no need to bind in order and
- more readable.

However, I see this is not aligned with the rest of the project, so I have no problem with converting to positional binding for where it is not yet used.

---

More notes:
- the test fails when changing the log level of the root logger and / or of the sqlserver QUERY class to `DEBUG`
- perf spec sometimes fails


core/src/main/scala/akka/persistence/r2dbc/internal/codec/TimestampCodec.scala
```diff
@@ -0,0 +1,73 @@
...
+    private val zone = TimeZone.getTimeZone("UTC").toZoneId
+
+    private def toInstant(timestamp: LocalDateTime) =
+      timestamp.atZone(zone).toInstant
+
+    override def decode(row: Row, name: String): Instant = toInstant(row.get(name, classOf[LocalDateTime]))
+
+    override def encode(timestamp: Instant): LocalDateTime = LocalDateTime.ofInstant(timestamp, zone)
+
+    override def now[T](): T = LocalDateTime.ofInstant(instantNow(), zone).asInstanceOf[T]
```
patriknw, line 58: that cast to `T` looks dangerous, for example this usage:
```
override def currentDbTimestamp(): Future[Instant] = Future.successful(timestampCodec.now())
```

wouldn't that cast a LocalDateTime to an Instant?

core/src/main/scala/akka/persistence/r2dbc/internal/codec/TimestampCodec.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  object H2TimestampCodec extends PostgresTimestampCodec
+
+  implicit class TimestampCodecRichStatement[T](val statement: Statement)(implicit codec: TimestampCodec)
+      extends AnyRef {
+    def bindTimestamp(name: String, timestamp: Instant): Statement = statement.bind(name, codec.encode(timestamp))
+    def bindTimestamp(index: Int, timestamp: Instant): Statement = statement.bind(index, codec.encode(timestamp))
+  }
+  implicit class TimestampCodecRichRow[T](val row: Row)(implicit codec: TimestampCodec) extends AnyRef {
+    def getTimestamp(index: String = "db_timestamp"): Instant = codec.decode(row, index)
```
patriknw, line 71: I think it is better to remove the default "db_timestamp" parameter value and be explicit about that from the call site. Aligns better with other row.get usages.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
```diff
@@ -153,15 +171,31 @@ private[r2dbc] class PostgresQueryDao(settings: R2dbcSettings, connectionFactory
...
+        row => row.getTimestamp())
       .map {
         case Some(time) => time
         case None       => throw new IllegalStateException(s"Expected one row for: $currentDbTimestampSql")
       }
   }
 
-  protected def tagsFromDb(row: Row, columnName: String): Set[String] =
-    setFromDb(row.get(columnName, classOf[Array[String]]))
+  //protected def tagsFromDb(row: Row, columnName: String): Set[String] = row.getTags(columnName)
```
patriknw, line 181: leftover?

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
```diff
@@ -228,6 +253,19 @@ private[r2dbc] class PostgresQueryDao(settings: R2dbcSettings, connectionFactory
     Source.futureSource(result.map(Source(_))).mapMaterializedValue(_ => NotUsed)
   }
 
+  protected def bindSelectBucketsSql(
+      stmt: Statement,
+      entityType: String,
+      fromTimestamp: Instant,
+      toTimestamp: Instant,
+      limit: Int): _root_.io.r2dbc.spi.Statement = {
```
patriknw, line 261: cleanup  `_root_` ...

patriknw (commented):
looking good

core/src/main/scala/akka/persistence/r2dbc/internal/codec/TimestampCodec.scala
sebastian-alfers: Yes, good idea. Same for the `getTags`, will change.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
sebastian-alfers: Yes!

core/src/main/scala/akka/persistence/r2dbc/internal/codec/TimestampCodec.scala
sebastian-alfers: Yes, right. I am fixing this now. It seems I can get rid of `def now[T](): T` completely, and fix the places where I am not yet binding a timestamp with the codec specific `.bindTimestamp`.

The unsafe casts must not yet have been triggered I assume.

core/src/main/scala/akka/persistence/r2dbc/internal/codec/TimestampCodec.scala
```diff
@@ -0,0 +1,68 @@
...
+ * INTERNAL API
+ */
+@InternalApi private[akka] sealed trait TimestampCodec {
+
+  def encode(timestamp: Instant): Any
+  def decode(row: Row, name: String): Instant
+  def decode(row: Row, index: Int): Instant
+
+  // should we name it just `now()`? The type should not be in the name...
+  def instantNow(): Instant = InstantFactory.now()
```
patriknw, line 27: I don't think we should have this at all in the TimestampCodec. It's not used by the rich statement and we can use `InstantFactory.now()` explicitly at the places where this is currently being used. The indirection via codec doesn't only adds confusion for creating an Instant.

core/src/main/scala/akka/persistence/r2dbc/internal/codec/TimestampCodec.scala
```diff
@@ -0,0 +1,68 @@
...
+
+    // should this come from config?
+    private val zone = TimeZone.getTimeZone("UTC").toZoneId
+
+    private def toInstant(timestamp: LocalDateTime) =
+      timestamp.atZone(zone).toInstant
+
+    override def decode(row: Row, name: String): Instant = toInstant(row.get(name, classOf[LocalDateTime]))
+
+    override def encode(timestamp: Instant): LocalDateTime = LocalDateTime.ofInstant(timestamp, zone)
```
patriknw, line 53: this is good, it wouldn't surprise me if r2dbc-sqlserver would add support for binding Instant in a future version. seems like a missing thing to me. anyway, for us this is good now

core/src/main/scala/akka/persistence/r2dbc/internal/codec/TimestampCodec.scala
```diff
@@ -0,0 +1,68 @@
...
+    override def decode(row: Row, index: Int): Instant = row.get(index, classOf[Instant])
+
+    override def encode(timestamp: Instant): Any = timestamp
+  }
+  object PostgresTimestampCodec extends PostgresTimestampCodec
+
+  case object SqlServerTimestampCodec extends TimestampCodec {
+
+    // should this come from config?
+    private val zone = TimeZone.getTimeZone("UTC").toZoneId
```
patriknw, line 46: slightly more direct:
```
private val zone = ZoneId.of("UTC")
```

core/src/main/scala/akka/persistence/r2dbc/internal/sqlserver/SqlServerQueryDao.scala
```diff
@@ -0,0 +1,207 @@
...
+      behindCurrentTime: FiniteDuration,
+      backtracking: Boolean,
+      minSlice: Int,
+      maxSlice: Int): String = {
+
+    def toDbTimestampParamCondition =
+      if (toDbTimestampParam) "AND db_timestamp <= @until" else ""
+
+    // we know this is a LocalDateTime, so the cast should be ok
+    def localNow: LocalDateTime = timestampCodec.encode(timestampCodec.instantNow()).asInstanceOf[LocalDateTime]
```
patriknw, line 117: This smells somewhat. Should we just skip the indirection to the codec here and create it by:
```
LocalDateTime.ofInstant(InstantFactory.now(), UTC)
```

core/src/main/scala/akka/persistence/r2dbc/internal/sqlserver/SqlServerQueryDao.scala
```diff
@@ -0,0 +1,207 @@
...
+
+    def behindCurrentTimeIntervalCondition =
+      if (behindCurrentTime > Duration.Zero)
+        s"AND db_timestamp < DATEADD(ms, -${behindCurrentTime.toMillis}, CAST('$localNow' as datetime2(6)))"
```
patriknw, line 121: Not possible to use bind parameter for the `localNow`, I guess?
For H2 and Postgres we use `CURRENT_TIMESTAMP` even though configured with app timestamp. `CURRENT_TIMESTAMP` or `SYSUTCDATETIME()` doesn't work here?

patriknw (commented):
looking good, a few comments about the timestamps

core/src/main/scala/akka/persistence/r2dbc/internal/codec/TimestampCodec.scala
sebastian-alfers: They discussed it already [here](https://github.com/r2dbc/r2dbc-mssql/issues/174) and [here](https://github.com/r2dbc/r2dbc-mssql/issues/278).

core/src/main/scala/akka/persistence/r2dbc/internal/sqlserver/SqlServerQueryDao.scala
sebastian-alfers: Seems to work 👍 I changed it and was running the test locally a few times with success.

This also allows the removal of the `def localNow: LocalDateTime` above, good improvement!

core/src/main/scala/akka/persistence/r2dbc/internal/codec/TimestampCodec.scala
sebastian-alfers: Changed it. Do you think the string `UTC` should come from config so it can be overridden by users?

patriknw: it should not be configurable, we should always use UTC

core/src/main/scala/akka/persistence/r2dbc/internal/sqlserver/SqlServerDurableStateDao.scala
```diff
@@ -0,0 +1,201 @@
...
+ */
+@InternalApi
+private[r2dbc] object SqlServerDurableStateDao {
+  private def log: Logger = LoggerFactory.getLogger(classOf[SqlServerDurableStateDao])
+}
+
+/**
+ * INTERNAL API
+ *
+ * Class for doing db interaction outside of an actor to avoid mistakes in future callbacks
```
johanandren, line 41: Apidoc seems off, it doesn't do anything with futures?
```suggestion
```

core/src/main/scala/akka/persistence/r2dbc/internal/sqlserver/SqlServerJournalDao.scala
```diff
@@ -0,0 +1,60 @@
...
+@InternalApi
+private[r2dbc] object SqlServerJournalDao {
+  private def log: Logger = LoggerFactory.getLogger(classOf[SqlServerJournalDao])
+  val TRUE = "1"
+}
+
+/**
+ * INTERNAL API
+ *
+ * Class for doing db interaction outside of an actor to avoid mistakes in future callbacks
```
johanandren, line 34:
```suggestion
```

core/src/main/scala/akka/persistence/r2dbc/internal/sqlserver/SqlServerQueryDao.scala
```diff
@@ -0,0 +1,206 @@
...
+@InternalApi
+private[r2dbc] object SqlServerQueryDao {
+  private def log: Logger = LoggerFactory.getLogger(classOf[SqlServerQueryDao])
+
+}
+
+/**
+ * INTERNAL API
+ *
+ * Class for doing db interaction outside of an actor to avoid mistakes in future callbacks
```
johanandren, line 42:
```suggestion
```

core/src/main/scala/akka/persistence/r2dbc/internal/PayloadCodec.scala
```diff
@@ -11,6 +11,8 @@ import io.r2dbc.postgresql.codec.Json
 import io.r2dbc.spi.Row
 import io.r2dbc.spi.Statement
 
+// move this file into internal/codec dir?
+
```
johanandren, line 15: Sounds reasonable, the JSON one as well in that case

core/src/main/scala/akka/persistence/r2dbc/internal/sqlserver/SqlServerSnapshotDao.scala
```diff
@@ -0,0 +1,212 @@
...
+ */
+@InternalApi
+private[r2dbc] object SqlServerSnapshotDao {
+  private def log: Logger = LoggerFactory.getLogger(classOf[SqlServerSnapshotDao])
+}
+
+/**
+ * INTERNAL API
+ *
+ * Class for doing db interaction outside of an actor to avoid mistakes in future callbacks
```
johanandren, line 41:
```suggestion
```

johanandren (commented):
Looking good, left some questions and nitpicks but I think the only big thing missing is documentation.

core/src/main/scala/akka/persistence/r2dbc/internal/codec/TagsCodec.scala
```diff
@@ -0,0 +1,83 @@
...
+import com.typesafe.config.Config
+import io.r2dbc.spi.Row
+import io.r2dbc.spi.Statement
+
+import akka.annotation.InternalApi
+
+/**
+ * INTERNAL API
+ */
+@InternalApi private[akka] sealed trait TagsCodec {
```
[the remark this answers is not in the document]
sebastian-alfers: Yes, I took that "codec" naming from the existing `PayloadCodec`. I have no strong opinion about naming that (or the one I introduce in this PR) differently, but let me look into what r2dbc provides from a codc perspective first.

core/src/main/scala/akka/persistence/r2dbc/internal/PayloadCodec.scala
sebastian-alfers: > the JSON one as well in that case

Moved the whole file, so assuming this is what you meant?!

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresJournalDao.scala
```diff
@@ -308,24 +319,34 @@ private[r2dbc] class PostgresJournalDao(journalSettings: R2dbcSettings, connecti
...
   override def deleteEventsTo(persistenceId: String, toSequenceNr: Long, resetSequenceNumber: Boolean): Future[Unit] = {
 
     def insertDeleteMarkerStmt(deleteMarkerSeqNr: Long, connection: Connection): Statement = {
+
+      // should we put this into its own class an re-use it where needed?
+      var i = 0
+      def getAndIncIndex(): Int = {
+        i += 1
+        i - 1
+      }
```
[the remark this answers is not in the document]
sebastian-alfers: Sounds like a good idea, going to refactor the existing usages of `getAndIncIndex` as well.

core/src/main/scala/akka/persistence/r2dbc/internal/sqlserver/SqlServerDialect.scala
```diff
@@ -0,0 +1,99 @@
...
+            .builder()
+            .option(ConnectionFactoryOptions.DRIVER, settings.driver)
+            .option(ConnectionFactoryOptions.HOST, settings.host)
+            .option(ConnectionFactoryOptions.PORT, Integer.valueOf(settings.port))
+            .option(ConnectionFactoryOptions.USER, settings.user)
+            .option(ConnectionFactoryOptions.PASSWORD, settings.password)
+            .option(ConnectionFactoryOptions.DATABASE, settings.database)
+            .option(ConnectionFactoryOptions.CONNECT_TIMEOUT, JDuration.ofMillis(settings.connectTimeout.toMillis))
+            //the option below is necessary to avoid https://github.com/r2dbc/r2dbc-mssql/issues/276
+            .option(MssqlConnectionFactoryProvider.PREFER_CURSORED_EXECUTION, false)
```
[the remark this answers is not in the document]
sebastian-alfers: Yes, lets put it at the end 👍 

While here, it seems I did not add all the "Supported ConnectionFactory Discovery Options" from [here](https://github.com/r2dbc/r2dbc-mssql?tab=readme-ov-file#getting-started).

ddl-scripts/create_tables_sqlserver.sql
```diff
@@ -0,0 +1,73 @@
+IF object_id('event_journal') is null
```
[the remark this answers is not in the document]
sebastian-alfers: There is also another option to query some internal table (that list existing object, like tables), but I think this version is shorter - and seems to work as well.

docker/docker-compose-sqlserver.yml
```diff
@@ -0,0 +1,10 @@
+version: '2.2'
+services:
+  sqlserver:
+    image: mcr.microsoft.com/mssql/server:2022-latest
+    container_name: sqlserver-db
+    environment:
+      - MSSQL_SA_PASSWORD=<YourStrong@Passw0rd>
+      - ACCEPT_EULA=Y
```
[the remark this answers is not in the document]
sebastian-alfers: The EULA this image links to is [this](http://go.microsoft.com/fwlink/?LinkId=746388) and it starts with the following (which makes me assume we are good):
 
> 	a.	Installation and Use.
		-	You may install and use up to ten copies of the software on your
premises to design, develop and test your programs for use with the software.

Also we use the same for jdbc.

migration-tests/src/test/scala/akka/persistence/r2dbc/migration/MigrationToolSpec.scala
```diff
@@ -74,6 +74,7 @@ class MigrationToolSpec
 
   private val migration = new MigrationTool(system)
 
+  // enable for sqlserver as well?
```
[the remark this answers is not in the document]
sebastian-alfers: Table creation would need to be tailored for sqlserver, skipping it for now.

core/src/main/scala/akka/persistence/r2dbc/internal/codec/TagsCodec.scala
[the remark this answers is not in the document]
sebastian-alfers: I don't really see how to make the r2dbc codec approach work with our way to choose the plugin a runtime via the `dialect` config. Maybe could work if we publish different artifacts per dialect (sbt submodules), so we can maintain custom `META-INF/services` per dialect - but I would rather not got for this in this PR.

docker/docker-compose-sqlserver.yml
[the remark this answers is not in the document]
johanandren: Thanks for double checking

ddl-scripts/create_tables_sqlserver.sql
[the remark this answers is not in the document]
johanandren: Yeah, no objection, just surprised it is so clunky.

core/src/main/scala/akka/persistence/r2dbc/internal/codec/TagsCodec.scala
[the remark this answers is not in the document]
johanandren: I think the codec API is not generic in R2DBC but one for each db plugin that supports pluggable codecs, so would be different SPI "provider configuration" files.

It's good enough that you took a look though, let's keep the current design and naming.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
```diff
@@ -93,6 +100,9 @@ private[r2dbc] class PostgresDurableStateDao(
     settings.connectionFactorySettings.poolSettings.closeCallsExceeding)(ec, system)
 
   private implicit val statePayloadCodec: PayloadCodec = settings.durableStatePayloadCodec
+  implicit val timestampCodec: TimestampCodec = settings.timestampCodec
+  private implicit val tagsCodec: TagsCodec = settings.tagsCodec
+  implicit val queryAdapter: QueryAdapter = settings.queryAdapter
```
johanandren, line 105: This and the `tiestampCodec` are just for internal access, right? `protect` or `private` em' to show intent

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
```diff
@@ -52,12 +55,16 @@ private[r2dbc] class PostgresQueryDao(settings: R2dbcSettings, connectionFactory
...
   import PostgresJournalDao.readMetadata
-  import PostgresQueryDao.setFromDb
 
   protected def log: Logger = PostgresQueryDao.log
-  protected val journalTable = settings.journalTableWithSchema
+  protected val journalTable: String = settings.journalTableWithSchema
   protected implicit val journalPayloadCodec: PayloadCodec = settings.journalPayloadCodec
+  protected implicit val timestampCodec: TimestampCodec = settings.timestampCodec
+  protected implicit val tagsCodec: TagsCodec = settings.tagsCodec
+  implicit val queryAdapter: QueryAdapter = settings.queryAdapter
```
johanandren, line 64: Another `private` or `protected`

johanandren (approved):
LGTM with a few more nitpicks. Would be good with a second LGTM before merging though.

patriknw (commented):
Looking good, only a few minor things.
Maybe it can be worth looking at the dialect in Akka Projections before we release to see that this fits in with that.

core/src/main/scala/akka/persistence/r2dbc/R2dbcSettings.scala
```diff
@@ -83,6 +88,18 @@ object R2dbcSettings {
 
     val connectionFactorySettings = ConnectionFactorySettings(config.getConfig("connection-factory"))
 
+    val (tagsCodec: TagsCodec, timestampCodec: TimestampCodec, queryAdapter: QueryAdapter) = {
```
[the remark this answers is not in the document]
sebastian-alfers: I think this `val`s are scoped within the method `apply` of that class, so it should be good.

ddl-scripts/create_tables_sqlserver.sql
```diff
@@ -0,0 +1,73 @@
...
+	  PRIMARY KEY(persistence_id, revision)
+	);
+
+-- `durable_state_slice_idx` is only needed if the slice based queries are used
+IF NOT EXISTS(SELECT * FROM sys.indexes WHERE name = 'durable_state_slice_idx' AND object_id = OBJECT_ID('durable_state'))
+    BEGIN
+        CREATE INDEX durable_state_slice_idx ON durable_state(slice, entity_type, db_timestamp, revision);
+    END;
+
+--DROP TABLE event_journal;
```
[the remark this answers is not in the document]
sebastian-alfers: No copy, it was my mistake. Table drops are in its own file.

docs/src/main/paradox/getting-started.md
```diff
@@ -99,6 +99,28 @@ akka.persistence.r2dbc.connection-factory = {
...
+akka.persistence.r2dbc.connection-factory = ${akka.persistence.r2dbc.sqlserver}
+akka.persistence.r2dbc.connection-factory = {
+  # overrides for default values from the 'akka.persistence.r2dbc.sqlserver' config block
+  user = "myuser"
+}
+```
+
+See @ref[Configuration](config.md) for more configuration details.
+
 ## Local testing with docker
```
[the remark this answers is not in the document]
sebastian-alfers: Do you think its fine if we start to advertise `docker compose` instead of `docker-compose`?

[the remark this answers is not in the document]
patriknw: I think that should be fine

core/src/main/scala/akka/persistence/r2dbc/internal/codec/QueryAdapter.scala
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (C) 2022 - 2023 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.persistence.r2dbc.internal.codec
+
+trait QueryAdapter {
```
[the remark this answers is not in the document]
sebastian-alfers: I can look into projections regarding sqlserver.

migration-tests/src/test/scala/akka/persistence/r2dbc/migration/MigrationToolSpec.scala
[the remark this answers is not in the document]
sebastian-alfers: Should I create a ticket for this?

sebastian-alfers:
> Maybe it can be worth looking at the dialect in Akka Projections before we release to see that this fits in with that.

I looked at projections and did not see anything missing, when it comes to integrating changes from this PR.

core/src/main/scala/akka/persistence/r2dbc/internal/codec/QueryAdapter.scala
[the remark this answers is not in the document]
sebastian-alfers: I was building the `sqlserver` dialect in projections based on a `publishLocal` snapshot from this branch, and did not see any problems in this regards.

migration-tests/src/test/scala/akka/persistence/r2dbc/migration/MigrationToolSpec.scala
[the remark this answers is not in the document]
patriknw: The migration tool is only using methods in the dao that we are using for other things already, if I'm not misremembering. Therefore it is not that important to test it with all dialects. I think the comment can be removed and nothing more.

patriknw (approved):
LGTM

migration-tests/src/test/scala/akka/persistence/r2dbc/migration/MigrationToolSpec.scala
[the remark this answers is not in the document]
patriknw: @sebastian-alfers I think I misunderstood. The question wasn't only about the test. The MigrationTool has it's own MigrationDao, which would be different for sqlserver. Please create an issue and we handle it if it comes up as a request.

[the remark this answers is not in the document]
sebastian-alfers: https://github.com/akka/akka-persistence-r2dbc/issues/509

Issue #504, opened 2024-01-18
Migration tool for Durable State


sebastian-alfers:
For a given list of persistence ids, run the migration tool for a jdbc based Durable State table.

PR #506, opened 2024-01-22
feat: Migration of Durable State, #504

patriknw:
* adding Durable State to the MigrationTool
* currently doesn't support retrieval of all persistenceIds because the jdbc plugin doesn't implement PersistenceIdsQuery, and jdbc plugin is currently the only other know plugin for durable state

References #504

### DIFF
--- a/migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
+++ b/migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
@@ -449,7 +449,7 @@ class MigrationTool(system: ActorSystem[_]) {
     val stateSerializer = serialization.findSerializerFor(stateAnyRef)
     val stateManifest = Serializers.manifestFor(stateSerializer, stateAnyRef)
 
-    // not possible to preserve timestamp
+    // not possible to preserve timestamp, because not included in GetObjectResult
     val timestamp = Instant.now()
 
     val serializedRow = SerializedStateRow(
@@ -460,22 +460,19 @@ class MigrationTool(system: ActorSystem[_]) {
       Some(serializedState),
       stateSerializer.identifier,
       stateManifest,
-      tags = Set.empty // not possible to preserve tags
+      tags = Set.empty // not possible to preserve tags, because not included in GetObjectResult
     )
     serializedRow
   }
 
   private def loadSourceDurableState(persistenceId: String, minRevision: Long): Future[Option[SelectedDurableState]] = {
-    if (sourceDurableStatePluginId == "")
-      Future.successful(None)
-    else
-      sourceDurableStateStore
-        .getObject(persistenceId)
-        .map {
-          case GetObjectResult(Some(value), revision) if revision >= minRevision =>
-            Some(SelectedDurableState(persistenceId, revision, value))
-          case _ => None
-        }
+    sourceDurableStateStore
+      .getObject(persistenceId)
+      .map {
+        case GetObjectResult(Some(value), revision) if revision >= minRevision =>
+          Some(SelectedDurableState(persistenceId, revision, value))
+        case _ => None
+      }
   }
 
 }
--- a/migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
+++ b/migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
@@ -5,12 +5,14 @@
 package akka.persistence.r2dbc.migration
 
 import java.time.Instant
+
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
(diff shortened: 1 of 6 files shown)


migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
```diff
@@ -345,4 +369,113 @@ class MigrationTool(system: ActorSystem[_]) {
...
+  }
+
+  private def serializedDurableStateRow(selectedDurableState: SelectedDurableState): SerializedStateRow = {
+    val stateAnyRef = selectedDurableState.value.asInstanceOf[AnyRef]
+    val serializedState = serialization.serialize(stateAnyRef).get
+    val stateSerializer = serialization.findSerializerFor(stateAnyRef)
+    val stateManifest = Serializers.manifestFor(stateSerializer, stateAnyRef)
+
+    // not possible to preserve timestamp
+    val timestamp = Instant.now()
```
johanandren, line 453: Why not, there surely is a timestamp in the JDBC store?

migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
```diff
@@ -345,4 +369,113 @@ class MigrationTool(system: ActorSystem[_]) {
...
+
+    val serializedRow = SerializedStateRow(
+      selectedDurableState.persistenceId,
+      selectedDurableState.revision,
+      timestamp,
+      timestamp,
+      Some(serializedState),
+      stateSerializer.identifier,
+      stateManifest,
+      tags = Set.empty // not possible to preserve tags
```
johanandren, line 463: Is that a separate table in JDBC or why?

migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
```diff
@@ -345,4 +369,113 @@ class MigrationTool(system: ActorSystem[_]) {
...
+    )
+    serializedRow
+  }
+
+  private def loadSourceDurableState(persistenceId: String, minRevision: Long): Future[Option[SelectedDurableState]] = {
+    if (sourceDurableStatePluginId == "")
+      Future.successful(None)
```
johanandren, line 470: Isn't that an error rather, misconfiguration?

migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
```diff
@@ -345,4 +369,113 @@ class MigrationTool(system: ActorSystem[_]) {
...
+    else
+      sourceDurableStateStore
+        .getObject(persistenceId)
+        .map {
+          case GetObjectResult(Some(value), revision) if revision >= minRevision =>
+            Some(SelectedDurableState(persistenceId, revision, value))
+          case _ => None
```
johanandren, line 477: Same here, if you pass it nonexistant ids, shouldn't that rather be a failure, or at least a logged warning?

johanandren (commented):
Lookin' good, some questions.

migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
patriknw: we use the the `DurableStateStore` api, so it's not tied to jdbc plugin specifically, and from that we don't get the timestamp

migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
patriknw: default value from reference.conf is
```
durable-state-plugin-id = "jdbc-durable-state-store"
```

migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
patriknw: I answered above question about timestamp, but same answer for tags

migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
patriknw: well, maybe this check is copy pasta from snapshot migration. I think it can be removed

migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
patriknw: since the migration tool is designed to be run while the source system could still be running, it is possible that it has been deleted

johanandren (approved):

PR #508, opened 2024-01-30
feat: Data partitions

patriknw:
Early draft, only multiple journal tables so far.

TODO:
- [x] Some tests are still failing and I will look into them. 
- [x] persistence ids queries that span over more than one data partition
- [ ] docs
- [ ] other FIXMEs

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -138,13 +138,14 @@ jobs:
 
       - name: Start DB
         run: |-
-          docker compose -f docker/docker-compose-postgres.yml up --wait
-          docker exec -i postgres-db psql -U postgres -t < ddl-scripts/create_tables_postgres_4.sql
+          docker compose -f docker/docker-compose-postgres-2.yml up --wait
+          docker exec -i postgres-db-0 psql -U postgres -t < ddl-scripts/create_tables_postgres_0-1.sql
+          docker exec -i postgres-db-1 psql -U postgres -t < ddl-scripts/create_tables_postgres_2-3.sql
 
       - name: sbt test
         run: |-
           cp .jvmopts-ci .jvmopts
-          sbt -Dakka.persistence.r2dbc.journal.table-data-partitions=4 test
+          sbt -Dconfig.resource=application-postgres-data-partitions.conf test
 
   test-yugabyte:
     name: Run tests with Yugabyte
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -112,6 +112,40 @@ jobs:
             -Dakka.persistence.r2dbc.state.payload-column-type=JSONB \
             "core/testOnly akka.persistence.r2dbc.PayloadSpec"
 
+  test-postgres-data-partitions:
+    name: Run test with Postgres and several data partitions
+    runs-on: ubuntu-22.04
+    if: github.repository == 'akka/akka-persistence-r2dbc'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3.1.0
+        with:
+          fetch-depth: 0
+
+      - name: Checkout GitHub merge
+        if: github.event.pull_request
+        run: |-
+          git fetch origin pull/${{ github.event.pull_request.number }}/merge:scratch
+          git checkout scratch
+
+      - name: Cache Coursier cache
+        uses: coursier/cache-action@v6.4.0
+
+      - name: Set up JDK 11
+        uses: coursier/setup-action@v1.3.0
+        with:
+          jvm: temurin:1.11.0
+
+      - name: Start DB
+        run: |-
+          docker compose -f docker/docker-compose-postgres.yml up --wait
+          docker exec -i postgres-db psql -U postgres -t < ddl-scripts/create_tables_postgres_4.sql
+
+      - name: sbt test
+        run: |-
+          cp .jvmopts-ci .jvmopts
+          sbt -Dakka.persistence.r2dbc.journal.table-data-partitions=4 test
+
   test-yugabyte:
     name: Run tests with Yugabyte
     runs-on: ubuntu-22.04
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -191,6 +191,9 @@ akka.persistence.r2dbc {
 # data partition 3.
 # This configuration cannot be changed in a rolling update, since the data must be moved
 # between the tables if number of data partitions is changed.
+# The number of Projection instances when using eventsBySlices must be greater than or equal
+# to the number of data partitions, because a query for a slice range cannot span over more
+# than one data partition.
 akka.persistence.r2dbc.data-partition {
   # How many tables the data will be partitioned over. The tables will have
   # the data partition as suffix, e.g. event_journal_0, event_journal_1.
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -7,18 +7,6 @@ akka.persistence.r2dbc {
     # name of the table to use for events
     table = "event_journal"
 
-    # Number of tables that the journal will be split into. The selection of data partition is
-    # made from the slice of the persistenceId. Must be between 1 and 1024 and a whole number
-    # divisor of 1024 (number of slices).
-    # For example, 4 data-partitions means that slice range (0 to 255) maps to data partition 0,
-    # (256 to 511) to data partition 1, (512 to 767) to data partition 3, and (768 to 1023) to
-    # data partition 3.
-    # The event_journal tables will have the data partition as suffix, e.g. _0, _1, _2, _3.
-    # When data-partitions is 1 there will only be one journal table, without suffix.
-    # This configuration cannot be changed in a rolling update, since the data must be moved
-    # between the tables if number of data partitions is changed.
-    table-data-partitions = 1
-
     # the column type to use for event payloads (BYTEA or JSONB)
     payload-column-type = "BYTEA"
 
@@ -195,6 +183,33 @@ akka.persistence.r2dbc {
   }
 }
 
+// #data-partition-settings
+# Number of tables and databases that the data will be split into. The selection of data
+# partition is made from the slice of the persistenceId.
+# For example, 4 data partitions means that slice range (0 to 255) maps to data partition 0,
+# (256 to 511) to data partition 1, (512 to 767) to data partition 3, and (768 to 1023) to
+# data partition 3.
+# This configuration cannot be changed in a rolling update, since the data must be moved
+# between the tables if number of data partitions is changed.
+akka.persistence.r2dbc.data-partition {
+  # How many tables the data will be partitioned over. The tables will have
+  # the data partition as suffix, e.g. event_journal_0, event_journal_1.
+  # Must be between 1 and 1024 and a whole number divisor of 1024 (number of slices).
+  # When number-of-partitions is 1 the table name is without suffix.
+  number-of-partitions = 1
+  # How many databases the tables will be partitioned over. A database corresponds to a connection
+  # factory with its own connection pool.
+  # Must be a whole number divisor of number-of-partitions, and less than or equal to number-of-partitions.
+  # For example, number-of-partitions=8 and number-of-databases=2 means that there will be a total of
+  # 8 tables in 2 databases, i.e. 4 tables in each database.
+  # The connection-factory setting will have the data partition range as suffix, e.g. with 8 data partitions and
+  # 2 databases the connection factory settings are connection-factory-0-3, connection-factory-4-7.
+  # When number-of-databases is 1 there will only be one connection factory, without suffix.
+  # number-of-databases > 1 not supported by H2.
+  number-of-databases = 1
+}
+// #data-partition-settings
+
 // #connection-settings
 akka.persistence.r2dbc {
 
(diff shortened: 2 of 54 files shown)


core/src/main/resources/reference.conf
```diff
@@ -182,6 +183,33 @@ akka.persistence.r2dbc {
...
+  # Must be a whole number divisor of number-of-partitions, and less than or equal to number-of-partitions.
+  # For example, number-of-partitions=8 and number-of-databases=2 means that there will be a total of
+  # 8 tables in 2 databases, i.e. 4 tables in each database.
+  # The connection-factory setting will have the data partition range as suffix, e.g. with 8 data partitions and
+  # 2 databases the connection factory settings are connection-factory-0-3, connection-factory-4-7.
+  # When number-of-databases is 1 there will only be one connection factory, without suffix.
+  # number-of-databases > 1 not supported by H2.
+  number-of-databases = 1
+}
+// #data-partition-settings
```
patriknw, line 211: @johanandren I have also implemented partitioning over multiple databases (several connection factories). Would be good if you can take a brief look before I continue with more than the journal, and other remaining tasks.

I hope these config comments explains how it's intended to work.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresJournalDao.scala
```diff
@@ -63,82 +63,77 @@ private[r2dbc] object PostgresJournalDao {
...
-      s"(SELECT db_timestamp + '1 microsecond'::interval FROM $journalTable " +
-      "WHERE persistence_id = ? AND seq_nr = ?)"
-
-    val insertEventWithParameterTimestampSql = {
-      if (journalSettings.dbTimestampMonotonicIncreasing)
-        sql"$baseSql ?) RETURNING db_timestamp"
-      else
-        sql"$baseSql GREATEST(?, $timestampSubSelect)) RETURNING db_timestamp"
-    }
+  protected def insertEventWithParameterTimestampSql(slice: Int): String = {
```
patriknw, line 82: pretty much all of the sql must be created for a given slice, which maps to a data partition, to resolve the table name with the data partition suffix

Might be a performance degradation with all string concatenation that has to be done for each request. Not sure if it matters? Maybe we could add some caching?

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
```diff
@@ -149,26 +151,41 @@ private[r2dbc] class PostgresQueryDao(settings: R2dbcSettings, connectionFactory
...
-  protected val allPersistenceIdsSql =
-    sql"SELECT DISTINCT(persistence_id) from $journalTable ORDER BY persistence_id LIMIT ?"
-
-  protected val persistenceIdsForEntityTypeSql =
-    sql"SELECT DISTINCT(persistence_id) from $journalTable WHERE persistence_id LIKE ? ORDER BY persistence_id LIMIT ?"
+  protected def allPersistenceIdsSql = {
+    // FIXME
+    require(
+      settings.numberOfDataPartitions == 1,
+      "allPersistenceIdsSql not implemented for more than one data-partition yet")
```
patriknw, line 158: these queries that are not for a specific slice would have to be implemented by several queries to each data partition and then merge the results

core/src/test/scala/akka/persistence/r2dbc/state/DurableStateUpdateWithChangeEventStoreSpec.scala
```diff
@@ -176,37 +178,36 @@ class DurableStateUpdateWithChangeEventStoreSpec
       envelopes.size shouldBe 2
     }
 
-  }
-
-  "publish change event" in {
```
patriknw, line 181: this was just misplaced, should be inside the `}`

.github/workflows/build-test.yml
```diff
@@ -112,6 +112,41 @@ jobs:
...
+      - name: Start DB
+        run: |-
+          docker compose -f docker/docker-compose-postgres-2.yml up --wait
+          docker exec -i postgres-db-0 psql -U postgres -t < ddl-scripts/create_tables_postgres_0-1.sql
+          docker exec -i postgres-db-1 psql -U postgres -t < ddl-scripts/create_tables_postgres_2-3.sql
+
+      - name: sbt test
+        run: |-
+          cp .jvmopts-ci .jvmopts
+          sbt -Dconfig.resource=application-postgres-data-partitions.conf test
```
patriknw, line 148: all tests pass when I run this locally

patriknw: and in CI

johanandren (commented):
Looks great so far, almost suspiciously simple.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
```diff
@@ -430,28 +412,38 @@ private[r2dbc] class PostgresQueryDao(settings: R2dbcSettings, executorProvider:
...
+                val stmt = connection.createStatement(persistenceIdsForEntityTypeAfterSql(sliceRange.min))
+                bindPersistenceIdsForEntityTypeAfterSql(stmt, entityType, likeStmtPostfix, after, actualLimit)
+
+              case None =>
+                val stmt = connection.createStatement(persistenceIdsForEntityTypeSql(sliceRange.min))
+                bindPersistenceIdsForEntityTypeSql(stmt, entityType, likeStmtPostfix, actualLimit)
+
+            },
+          row => row.get("persistence_id", classOf[String]))
+      }
```
johanandren, line 434: So limit will now in fact potentially yield `db partion * table partition * limit` entries, worth mentioning somewhere?

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
```diff
@@ -430,28 +412,38 @@ private[r2dbc] class PostgresQueryDao(settings: R2dbcSettings, executorProvider:
...
-      connection =>
-        afterId match {
-          case Some(after) =>
-            val stmt = connection.createStatement(persistenceIdsForEntityTypeAfterSql)
-            bindPersistenceIdsForEntityTypeAfterSql(stmt, entityType, likeStmtPostfix, after, limit)
-
-          case None =>
-            val stmt = connection.createStatement(persistenceIdsForEntityTypeSql)
-            bindPersistenceIdsForEntityTypeSql(stmt, entityType, likeStmtPostfix, limit)
+    val results: immutable.IndexedSeq[Future[immutable.IndexedSeq[String]]] =
```
johanandren, line 417: Would it be better to query one partition at a time and stream out those results than to load all results from all partitions? Maybe fine since only persistence ids?

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
```diff
@@ -461,27 +453,37 @@ private[r2dbc] class PostgresQueryDao(settings: R2dbcSettings, executorProvider:
...
+                val stmt = connection.createStatement(allPersistenceIdsAfterSql(sliceRange.min))
+                bindAllPersistenceIdsAfterSql(stmt, after, actualLimit)
+
+              case None =>
+                connection
+                  .createStatement(allPersistenceIdsSql(sliceRange.min))
+                  .bind(0, actualLimit)
+            },
+          row => row.get("persistence_id", classOf[String]))
+      }
```
johanandren, line 474: Maybe the expectation is that the source is entries when `persistenceIds` call was made rather than the stream materialization, but this could be surprinsing, should probably be triggered by the futureSource rather than just collected there?

core/src/main/scala/akka/persistence/r2dbc/internal/R2dbcExecutor.scala
```diff
@@ -396,17 +397,24 @@ class R2dbcExecutor(
...
+        val connectionFactory = connectionFactoryProvider.connectionFactoryFor(connectionFactoryConfigPath)
+        val executor = new R2dbcExecutor(
+          connectionFactory,
+          log,
+          settings.logDbCallsExceeding,
+          settings.connectionFactorySettings.poolSettings.closeCallsExceeding)
+        // it's just a cache so no need for guarding concurrent updates or visibility
+        cache = cache.updated(slice, executor)
+        executor
+    }
```
johanandren, line 417: 👍

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
patriknw: I think the result must be sorted, for the paged queries, and that is used also by the ordinary `currentPersistenceIds` which is using a ContinuousQuery to fetch all in chunks.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
patriknw: It's table partitions * limit. data partion = table, then those tables can optionally be spread over multiple databases, but there will not be a multipler

Yeah, I know it's not optimal to query all partitions for the same thing and then sort and truncate the result, but as far as I can see it that's the only way to do it. Ordering and paging as mentioned in other comment.

patriknw: and these persistenceIds queries are intended for backoffice, not heavy use

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
patriknw: I think that is the same for everything here, e.g. `eventsByPersistenceId`. So if it's better with lazy query we should change that in a separate pr.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
johanandren: Ok, combined with the fact that the partitioning is very specific performance trixery, maybe it is fine.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
johanandren: Yeah, maybe that is anyway the expectation, let's leave this and the others as is.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
johanandren: Didn't think of the ordering, yes, I can't see any other way than this given that.

patriknw: Actually, for the `currentPersistenceIds()` which retrieves all in chunks, it's probably possible to concat several of those queries where each go to a separate data partition. Would result in a different total order than we have today, but that might be fine. However, the other queries with user defined paging key can't use that, and therefore I prefer to keep them the same.

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/R2dbcDurableStateStore.scala
```diff
@@ -65,20 +66,22 @@ class R2dbcDurableStateStore[A](system: ExtendedActorSystem, config: Config, cfg
...
   log.debug("R2DBC journal starting up with dialect [{}]", settings.dialectName)
 
   private val typedSystem = system.toTyped
   private val serialization = SerializationExtension(system)
   private val persistenceExt = Persistence(system)
-  private val stateDao = settings.connectionFactorySettings.dialect.createDurableStateDao(
-    settings,
-    ConnectionFactoryProvider(typedSystem)
-      .connectionFactoryFor(sharedConfigPath + ".connection-factory"))(typedSystem)
+  // FIXME maybe this is using the wrong executionContext, H2Dialect is using another dispatcher?
```
patriknw, line 74: this might be a regression, that would be important to follow up on before making a real release

patriknw:
@johanandren would be ok that we merge this as first step and I continue in additional PR? Would like to cut a milestone?

octonato (approved):
LGTM

It's a little unfortunate that some queries need to go over different databases and tables and merge results (and fetch and drop data), but they are more edge cases. 

The main use case (write events and read events) will benefit from the partition and that's what matters.

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/R2dbcDurableStateStore.scala
johanandren: Not sure if this comment is related to that, but, as far as I remember H2 needed to use a blocking dispatcher because it synchronizes and filesystem stuff on the calling thread.

patriknw: yes, that is what I saw, and I think this change makes it use the default

johanandren (approved):
Sounds good merging this and continuing on top

patriknw:
hm, ci not triggering

patriknw:
I'll create separate PRs with the remaining work

core/src/main/scala/akka/persistence/r2dbc/state/scaladsl/R2dbcDurableStateStore.scala
patriknw: fixed in https://github.com/akka/akka-persistence-r2dbc/pull/518

pvlugter:
👍🏼 great to see this in, and nice and simple, after being discussed so many times.

PR #514, opened 2024-02-02
chore: Make Sql.Interpolation backwards compatible

patriknw:
Follow up to https://github.com/akka/akka-persistence-r2dbc/pull/505
`Sql.Interpolation` is used in other projects that we don't want to break.

Also a cleanup of the codec settings.

### DIFF
--- a/core/src/main/mima-filters/1.2.1.backwards.excludes/mssql-dialect.excludes
+++ b/core/src/main/mima-filters/1.2.1.backwards.excludes/mssql-dialect.excludes
@@ -1,8 +1,4 @@
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.persistence.r2dbc.R2dbcSettings.this")
-ProblemFilters.exclude[MissingClassProblem]("akka.persistence.r2dbc.internal.Sql$Interpolation$")
-ProblemFilters.exclude[DirectMissingMethodProblem]("akka.persistence.r2dbc.internal.Sql.Interpolation")
-ProblemFilters.exclude[DirectMissingMethodProblem]("akka.persistence.r2dbc.internal.Sql.Interpolation")
-ProblemFilters.exclude[DirectMissingMethodProblem]("akka.persistence.r2dbc.internal.Sql#Interpolation.this")
 ProblemFilters.exclude[MissingClassProblem]("akka.persistence.r2dbc.internal.h2.H2Utils")
 ProblemFilters.exclude[MissingClassProblem]("akka.persistence.r2dbc.internal.h2.H2Utils$")
 ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.persistence.r2dbc.R2dbcSettings.journalPayloadCodec")
--- a/core/src/main/scala/akka/persistence/r2dbc/R2dbcSettings.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/R2dbcSettings.scala
@@ -54,20 +54,11 @@ object R2dbcSettings {
           s"Expected akka.persistence.r2dbc.$prefix.payload-column-type to be one of 'BYTEA', 'JSON' or 'JSONB' but found '$t'")
     }
 
-    val journalPayloadCodec: PayloadCodec =
-      if (useJsonPayload("journal")) PayloadCodec.JsonCodec else PayloadCodec.ByteArrayCodec
-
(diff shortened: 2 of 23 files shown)


core/src/main/mima-filters/1.2.1.backwards.excludes/mssql-dialect.excludes
```diff
@@ -1,8 +1,4 @@
 ProblemFilters.exclude[DirectMissingMethodProblem]("akka.persistence.r2dbc.R2dbcSettings.this")
```
patriknw, line 1: the file should be latest release, 1.2.1, not 1.2.2

core/src/main/scala/akka/persistence/r2dbc/internal/Sql.scala
```diff
@@ -21,7 +23,19 @@ object Sql {
...
+      fillInParameterNumbers(trimLineBreaks(sc.s(args: _*)))
+  }
+
+  /**
+   * INTERNAL API: Scala string interpolation with `sql` prefix. Replaces `?` with numbered `\$1`, `\$2` for bind
+   * parameters. Use `??` to include a literal `?`. Trims whitespace, including line breaks. Standard string
+   * interpolation arguments `$` can be used.
+   */
+  @InternalApi private[akka] implicit class InterpolationWithAdapter(val sc: StringContext)(implicit
+      adapter: QueryAdapter)
```
patriknw, line 37: `QueryAdapter` is `private[akka]` and therefore I think the `InterpolationWithAdapter` should also be `private[akka]`, otherwise we should also make QueryAdapter public (InternalStableApi)

Note that this is `AnyRef` because of two parameters, instead of `AnyVal`. I'm not sure this idea with the adapter is worth the indirect complexity it adds. Doesn't we override all sql for sqlserver anyway in the SqlServerDao classes? Those could use `@` as the placeholder instead of `?` and then we could change the ordinary `fillInParameterNumbers` to look for `?` and `@` and replace accordingly. I'll might try that in a separate pr.

patriknw: ok, we don't override all sql, which is good. I'll think a little more about this, but this PR can be merged seperately

core/src/main/scala/akka/persistence/r2dbc/R2dbcSettings.scala
```diff
@@ -328,6 +324,39 @@ final class PublishEventsDynamicSettings(config: Config) {
...
+final class CodecSettings(
+    val journalPayloadCodec: PayloadCodec,
+    val snapshotPayloadCodec: PayloadCodec,
+    val durableStatePayloadCodec: PayloadCodec,
+    val tagsCodec: TagsCodec,
+    val timestampCodec: TimestampCodec,
+    val queryAdapter: QueryAdapter) {
+
+  // implicits that can be imported
+  object JournalImplicits {
```
sebastian-alfers, line 340: Have not seen this very often, objects within a case class. But I guess it is fine. Alternative could be something based on inheritance or so, this this `CodecSettings.this.journalPayloadCodec` could be avoided.

core/src/main/scala/akka/persistence/r2dbc/internal/Sql.scala
sebastian-alfers: Correct, we do not override all sql statements, which is why this QueryAdaper was introduced.

johanandren (approved):

sebastian-alfers (approved):
Good improvement to avoid having to import each coded individually.

core/src/main/scala/akka/persistence/r2dbc/internal/Sql.scala
patriknw: I leave it as this for now. Projections will still be able to use QueryAdapter since it's `private[akka]`.

core/src/main/scala/akka/persistence/r2dbc/R2dbcSettings.scala
patriknw: CodecSettings.this.journalPayloadCodec was just because I used the same names

patriknw: yeah, the object is only for name scoping

PR #521, opened 2024-02-09
build: Update actions due to Node.js 16 deprecation

patriknw:
* Node.js 16 actions are deprecated
* actions/checkout, coursier/cache-action, coursier/setup-action
* using same versions as in akka-projection


johanandren (approved):
LGTM

BTW I think @ennru was building a list of maintenance tasks like this across all Akka repos.

PR #522, opened 2024-02-09
perf: Cache sql construction

patriknw:
* avoid string concatenation and allocations
* since they are  now created each time its used due to the data partitions by slice

@johanandren can you sanity check this before I continue?

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/Sql.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/Sql.scala
@@ -5,6 +5,7 @@
 package akka.persistence.r2dbc.internal
 
 import scala.annotation.varargs
+import scala.collection.immutable.IntMap
 
 import akka.annotation.InternalApi
 import akka.annotation.InternalStableApi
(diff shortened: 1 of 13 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/Sql.scala
```diff
@@ -81,4 +82,29 @@ object Sql {
     }
   }
 
+  final class Cache {
```
sebastian-alfers, line 85: There is also this `mutable.Map.getOrElseUpdate` from [here](https://www.scala-lang.org/api/2.13.7/scala/collection/mutable/Map.html#getOrElseUpdate(key:K,op:=%3EV):V) - maybe this is more convenient, but I am not sure about the performance implications.

patriknw: I think we can use immutable here since there will be no updates after initial population

johanandren (commented):
Looks like a good idea

patriknw:
Completed this with caching in all places (where it makes sense)

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
```diff
@@ -107,24 +110,26 @@ private[r2dbc] class PostgresDurableStateDao(executorProvider: R2dbcExecutorProv
...
   }
 
-  protected def selectStateSql(slice: Int, entityType: String): String = {
-    val stateTable = settings.getDurableStateTableWithSchema(entityType, slice)
-    sql"""
-    SELECT revision, state_ser_id, state_ser_manifest, state_payload, db_timestamp
-    FROM $stateTable WHERE persistence_id = ?"""
-  }
+  protected def selectStateSql(slice: Int, entityType: String): String =
+    sqlCache.get(slice, s"selectStateSql-$entityType") {
```
patriknw, line 114: when reviewing this, the highest risk of mistake (by me) is using the wrong cache key, such as accidentally using the same in several places

johanandren (approved):
LGTM, but way less sure it is a good idea at all after reviewing it. Esp methods that create a string with slice numbers in them in each call to avoid creating another string.

patriknw:
> Esp methods that create a string with slice numbers in them in each call to avoid creating another string.

yeah, some might be questionable, but some might actually do much more than it looks, like those that create the slice ranges `IN (0, 1, 2, ..., 255)`

PR #534, opened 2024-03-20
fix: use val for logging in all SqlServer daos

sebastian-alfers:
As discovered here https://github.com/akka/akka-projection/pull/1122#discussion_r1531862868


patriknw (approved):
LGTM

PR #544, opened 2024-04-04
docs: clarify migration and retrieval of durable state based persistence ids

sebastian-alfers:


docs/src/main/paradox/migration.md
```diff
@@ -9,6 +9,8 @@ idempotent result. Full rolling update when switching database or Persistence pl
 you can migrate most of the data while the system is online and then have a short full shutdown while
 migrating the remaining data that was written after the previous online migration.
 
+The migration tool is intended to run as a separate, standalone application and should not be part of the same jvm process the main application is running under.
```
patriknw, line 12:
```suggestion
The migration tool is intended to run as a separate, standalone application and should not be part of the same jvm process as the main application is running under.
```

docs/src/main/paradox/migration.md
```diff
@@ -74,6 +76,8 @@ The migration tool can be run as main class `akka.persistence.r2dbc.migration.Mi
 
 Durable State is not migrated by `MigrationTool.migrateAll`, instead you need to use `MigrationTool.migrateDurableStates` for a given list of persistence ids.
 
+This list can be retrieved by using `currentPersistenceIds` of @ref:[Durable state queries](./query.md#durable-state-queries).
```
patriknw, line 79: This is misleading because you would have to retrieve them from the source (old) database, and akka-persistence-jdbc doesn't implement that query. Isn't it obvious that you can retrieve such list of persistence ids with just sql select?

sebastian-alfers: You are right, I have to remove that sentence as it was a misunderstanding.

docs/src/main/paradox/migration.md
```diff
@@ -9,6 +9,8 @@ idempotent result. Full rolling update when switching database or Persistence pl
 you can migrate most of the data while the system is online and then have a short full shutdown while
 migrating the remaining data that was written after the previous online migration.
 
+The migration tool is intended to run as a separate, standalone application and should not be part of the same jvm process as the main application is running under.
```
patriknw, line 12: @sebastian-alfers Is this PR ready?

sebastian-alfers: Yes.

patriknw (approved):

patriknw:
Flaky SQL Server test, but ignoring that since this is clearly just a doc addition

Issue #552, opened 2024-04-18
migration broken for updated state


sebastian-alfers:
The current logic for storing and updating Durable State decides based on revision=1 to `INSERT`, otherwise to `UPDATE`. 

This logic works for a production application but not for a migration where the state could come already with an updated revision which would cause an `UPDATE` query on the new table - which fails since there is no state to update.

PR #553, opened 2024-04-18
fix: insert instead of update state with updated revision during migration

sebastian-alfers:
Fixes https://github.com/akka/akka-persistence-r2dbc/issues/552

---

Very pragmatic approach to fix this issue. I first tried to add another path to make a true upsert during migration, but this caused the complexity of the `PostgresDurableStateDao` to increase even further. So I tried this version instead, and it seems to work ™️

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/R2dbcSettings.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/R2dbcSettings.scala
@@ -73,6 +73,7 @@ object R2dbcSettings {
       configToMap(config.getConfig("state.change-handler"))
 
     val durableStateAssertSingleWriter: Boolean = config.getBoolean("state.assert-single-writer")
+    val migration: Boolean = config.hasPath("migration")
 
     val numberOfDataPartitions = config.getInt("data-partition.number-of-partitions")
     require(
@@ -173,6 +174,7 @@ object R2dbcSettings {
       snapshotsTable,
       durableStateTable,
       durableStateAssertSingleWriter,
+      migration,
       logDbCallsExceeding,
       querySettings,
       dbTimestampMonotonicIncreasing,
@@ -224,6 +226,7 @@ final class R2dbcSettings private (
     val snapshotsTable: String,
     val durableStateTable: String,
     val durableStateAssertSingleWriter: Boolean,
+    val migration: Boolean,
     val logDbCallsExceeding: FiniteDuration,
     val querySettings: QuerySettings,
     val dbTimestampMonotonicIncreasing: Boolean,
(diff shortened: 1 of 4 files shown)


core/src/main/scala/akka/persistence/r2dbc/R2dbcSettings.scala
```diff
@@ -73,6 +73,7 @@ object R2dbcSettings {
       configToMap(config.getConfig("state.change-handler"))
 
     val durableStateAssertSingleWriter: Boolean = config.getBoolean("state.assert-single-writer")
+    val migration: Boolean = config.hasPath("migration")
```
patriknw, line 76: I don't think this should be the way to decide "migration" mode. They may have `akka-persistence-r2dbc-migration` in classpath and accidentally run in migration mode for the real app.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
```diff
@@ -506,13 +507,18 @@ private[r2dbc] class PostgresDurableStateDao(executorProvider: R2dbcExecutorProv
...
-    result.map { case (updatedRows, changeEventTimestamp) =>
-      if (updatedRows != 1)
-        throw new IllegalStateException(
-          s"Update failed: durable state for persistence id [${state.persistenceId}] could not be updated to revision [${state.revision}]")
-      else {
+    result.flatMap { case (updatedRows, changeEventTimestamp) =>
+      if (updatedRows != 1) {
+        if (settings.migration) {
+          // during migration,
+          upsertState(state, value, changeEvent, insertRevisionOverride = true)
```
patriknw, line 514: I think it would be more clean to have a separate dao method that is used from migration. It can do a real upsert instead of the insert/update attempts. It would not have to deal with changeEvent and it should not use the assert-single-writer check.

sebastian-alfers:
Looking for a different approach here.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
sebastian-alfers: A real upsert does not work either, since the `ON CONFLICT(persistence_id,revision)` only works on exact matches to decide if to insert or to update.

In our case, want to update if there is row with the same persistence id and a an `existing revision = new revision-1`

patriknw: why do we need to include the revision in the upsert condition? can't we use `ON CONFLICT(persistence_id)`?

> existing revision = new revision-1

I don't think that is true for migration, and the reason why the instructions were to disable the single-writer assert.

sebastian-alfers: For a `ON CONFLICT(persistence_id)` definition, I got:

```
io.r2dbc.postgresql.ExceptionFactory$PostgresqlBadGrammarException, with message: there is no unique or exclusion constraint matching the ON CONFLICT specification. (MigrationToolSpec.scala:416)
```

sebastian-alfers: But deciding on a insert / update should be doable with a simple upfront query, with minimal refactoring of the existing code in `PostgresDurableStateDao`.

PR #554, opened 2024-04-19
fix: insert instead of update new state with updated revision during migration

sebastian-alfers:
Fixes https://github.com/akka/akka-persistence-r2dbc/issues/552, successor of https://github.com/akka/akka-persistence-r2dbc/pull/553.

---

Before this pr, a new migrated state with an evolved revision was attempted to be `UPDATE`ed which failed as there exists no state to be upgraded.

This pr
- Applies an `INSERT` if a state is migrated that does not have an existing revision of its persistence id.
- Applies an `UPDATE` if a state is migrated that does have an existing revision of its persistence id.

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
@@ -89,14 +89,14 @@ private[r2dbc] class PostgresDurableStateDao(executorProvider: R2dbcExecutorProv
   import settings.codecSettings.DurableStateImplicits._
   protected def log: Logger = PostgresDurableStateDao.log
 
-  private val persistenceExt = Persistence(system)
+  protected val persistenceExt = Persistence(system)
 
   private val sqlCache = Sql.Cache(settings.numberOfDataPartitions > 1)
 
   // used for change events
   private lazy val journalDao: JournalDao = dialect.createJournalDao(executorProvider)
 
-  private lazy val additionalColumns: Map[String, immutable.IndexedSeq[AdditionalColumn[Any, Any]]] = {
+  protected lazy val additionalColumns: Map[String, immutable.IndexedSeq[AdditionalColumn[Any, Any]]] = {
     settings.durableStateAdditionalColumnClasses.map { case (entityType, columnClasses) =>
       val instances = columnClasses.map(fqcn => AdditionalColumnFactory.create(system, fqcn))
       entityType -> instances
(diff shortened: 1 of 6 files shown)


migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
```diff
@@ -123,6 +123,12 @@ class MigrationTool(system: ActorSystem[_]) {
   private val targetPluginId = migrationConfig.getString("target.persistence-plugin-id")
   private val targetR2dbcSettings = R2dbcSettings(system.settings.config.getConfig(targetPluginId))
 
+  targetR2dbcSettings.dialectName
+  require(
+    !targetR2dbcSettings.durableStateAssertSingleWriter,
+    "When running the MigrationTool the " +
```
johanandren, line 129: "When" vs "While"? (Once migration is done it can be set to on, right?)

migration/src/main/scala/akka/persistence/r2dbc/migration/DurableStateMigrationToolDao.scala
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  private def insert(
+      r2dbcExecutor: R2dbcExecutor,
+      newRow: SerializedStateRow,
+      newValue: Any,
+      slice: Int,
+      entityType: String) = {
+    r2dbcExecutor
+      .updateOne(s"update state [${newRow.persistenceId}] for migration") { connection =>
```
johanandren, line 47:
```suggestion
      .updateOne(s"insert state [${newRow.persistenceId}] for migration") { connection =>
```

migration/src/main/scala/akka/persistence/r2dbc/migration/DurableStateMigrationToolDao.scala
```diff
@@ -0,0 +1,66 @@
...
+    extends PostgresDurableStateDao(executorProvider, dialect) {
+
+  def upsertDurableState(state: SerializedStateRow, value: Any): Future[Done] = {
+    val slice = persistenceExt.sliceForPersistenceId(state.persistenceId)
+    val r2dbcExecutor = executorProvider.executorFor(slice)
+    val entityType = PersistenceId.extractEntityType(state.persistenceId)
+    readState(state.persistenceId).flatMap {
+      case Some(existingState) => update(r2dbcExecutor, existingState, state, value, slice, entityType)
+      case None                => insert(r2dbcExecutor, state, value, slice, entityType)
+    }
```
johanandren, line 37: Wouldn't it be better to leverage the upsert support in Postgres, instead of not having to do a read roundtrip for every migrated state write?

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
```diff
@@ -89,14 +89,14 @@ private[r2dbc] class PostgresDurableStateDao(executorProvider: R2dbcExecutorProv
...
-  private val persistenceExt = Persistence(system)
+  protected val persistenceExt = Persistence(system)
 
   private val sqlCache = Sql.Cache(settings.numberOfDataPartitions > 1)
 
   // used for change events
   private lazy val journalDao: JournalDao = dialect.createJournalDao(executorProvider)
 
-  private lazy val additionalColumns: Map[String, immutable.IndexedSeq[AdditionalColumn[Any, Any]]] = {
+  protected lazy val additionalColumns: Map[String, immutable.IndexedSeq[AdditionalColumn[Any, Any]]] = {
```
johanandren, line 99: IntelliJ says it could be kept private?

migration/src/main/scala/akka/persistence/r2dbc/migration/DurableStateMigrationToolDao.scala
sebastian-alfers: Yes, I had the same idea and Patrik [suggested](https://github.com/akka/akka-persistence-r2dbc/pull/553) the same. But when I looked at it, it seemed that an upsert can only decide on the exact primary key match, which is not what we want here I think.

migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
sebastian-alfers: 👍

migration/src/main/scala/akka/persistence/r2dbc/migration/DurableStateMigrationToolDao.scala
johanandren: I see, we have both pid and revision as part of primary key, I didn't remember that. Probably no way around it then. 👍

patriknw: wow, `PRIMARY KEY(persistence_id, revision)` must be an oversight. We don't ever want to store more than one row for a persistence_id. The application logic takes care of it, but feels wrong. Created ticket https://github.com/akka/akka-persistence-r2dbc/issues/558 for separate follow up.

patriknw (approved):
LGTM

johanandren:
I agree, that sounds better to me @sebastian-alfers (if we have time to spend on it that is)

sebastian-alfers:
Closing this to rework for real upsert since https://github.com/akka/akka-persistence-r2dbc/pull/559 is about to get merged.

PR #559, opened 2024-04-22
chore: remove revision from Durable State primary key

### LINKED ISSUE #558: revision shouldn't be part of durable_state pk
```
PRIMARY KEY(persistence_id, revision)
```

We don't ever want to store more than one row for a persistence_id. The application logic takes care of it, but would be good to have a correct primary key.

sebastian-alfers:
Fixes #558


johanandren (approved):
LGTM, seems safe to me.

patriknw (approved):
LGTM

PR #560, opened 2024-04-23
fix: upsert for Durable State to fix updated revision during migration

sebastian-alfers:
Has to be rebased on https://github.com/akka/akka-persistence-r2dbc/pull/559.

I noticed that additional columns are currently not supported.


migration/src/main/scala/akka/persistence/r2dbc/migration/DurableStateMigrationToolDao.scala
```diff
@@ -0,0 +1,79 @@
...
+  def upsertDurableState(state: SerializedStateRow, value: Any): Future[Done] = {
+    val slice = persistenceExt.sliceForPersistenceId(state.persistenceId)
+
+    val r2dbcExecutor = executorProvider.executorFor(slice)
+    val entityType = PersistenceId.extractEntityType(state.persistenceId)
+    val stateTable = settings.getDurableStateTableWithSchema(entityType, slice)
+
+    //val additionalBindings = getAdditionalBindings(state, value, entityType)
+    //val additionalCols = additionalInsertColumns(additionalBindings)
+    //val additionalParams = additionalInsertParameters(additionalBindings)
```
patriknw, line 42: these can be removed, not used by migration tool

migration/src/main/scala/akka/persistence/r2dbc/migration/MigrationTool.scala
```diff
@@ -123,6 +123,12 @@ class MigrationTool(system: ActorSystem[_]) {
   private val targetPluginId = migrationConfig.getString("target.persistence-plugin-id")
   private val targetR2dbcSettings = R2dbcSettings(system.settings.config.getConfig(targetPluginId))
 
+  targetR2dbcSettings.dialectName
+  require(
+    !targetR2dbcSettings.durableStateAssertSingleWriter,
+    "While running the MigrationTool the " +
+    "`akka.persistence.r2dbc.state.assert-single-writer` configuration must be set to off.")
```
patriknw, line 130: This shouldn't be necessary any more, and the comment about that in docs can be removed.

patriknw (commented):
looking good

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
```diff
@@ -375,75 +375,45 @@ private[r2dbc] class PostgresDurableStateDao(executorProvider: R2dbcExecutorProv
...
-          val slice = persistenceExt.sliceForPersistenceId(state.persistenceId)
-          val upsert = AdditionalColumn.Upsert(state.persistenceId, entityType, slice, state.revision, value)
-          columns.map(c => EvaluatedAdditionalColumnBindings(c, c.bind(upsert)))
-      }
 
       if (state.revision == 1) {
         def insertStatement(connection: Connection): Statement = {
           val stmt = connection
-            .createStatement(insertStateSql(slice, entityType, additionalBindings))
+            .createStatement(sql"${insertStateSql(slice, entityType, bindings)}")
```
patriknw, line 390: why is the `sql` moved here?

migration/src/main/scala/akka/persistence/r2dbc/migration/DurableStateMigrationToolDao.scala
```diff
@@ -0,0 +1,99 @@
...
+         DO UPDATE SET
+           slice = EXCLUDED.slice,
+           entity_type = EXCLUDED.entity_type,
+           revision = EXCLUDED.revision,
+           state_ser_id = EXCLUDED.state_ser_id,
+           state_ser_manifest = EXCLUDED.state_ser_manifest,
+           state_payload = EXCLUDED.state_payload,
+           tags = EXCLUDED.tags,
+           test_column = EXCLUDED.test_column,
+           db_timestamp = EXCLUDED.db_timestamp
```
patriknw, line 60: How does this work with additionalColumns? they are bound to the insertStateSql, but not part of those EXCLUDED updates.

If that is true I think it's a good indication that this is becoming too complex with risk of bugs and more difficult to maintain, and we should instead go back to #554

sebastian-alfers: Right, additional Columns (as well as change handler?) are missing here.

sebastian-alfers:
Closing this case to go back to https://github.com/akka/akka-persistence-r2dbc/pull/554 to avoid increased complexity.

PR #561, opened 2024-04-24
chore: drop revision from Durable State by slice index and migration docs

sebastian-alfers:
Mentioned in https://github.com/akka/akka-persistence-r2dbc/issues/558#issuecomment-2071979388

### DIFF
--- a/docs/src/main/paradox/migration-guide.md
+++ b/docs/src/main/paradox/migration-guide.md
@@ -14,7 +14,7 @@ DROP CONSTRAINT durable_state_pkey,
 ADD PRIMARY KEY(persistence_id);
 ```
 
-Same goes for the (optional) index `` used for slice base queries. To remove the column from the index, use:
+Same goes for the (optional) index `durable_state_slice_idx` used for slice base queries. To remove the column from the index, use:
 
 Postgres / Yugabyte:
 : ```sql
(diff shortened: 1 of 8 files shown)


docs/src/main/paradox/migration-guide.md
```diff
@@ -14,6 +14,18 @@ DROP CONSTRAINT durable_state_pkey,
 ADD PRIMARY KEY(persistence_id);
 ```
 
+Same goes for the (optional) index `` used for slice base queries. To remove the column from the index, use:
+
+Postgres / Yugabyte:
+: ```sql
+DROP INDEX IF EXISTS durable_state_slice_idx;
+CREATE INDEX IF NOT EXISTS durable_state_slice_idx ON durable_state(slice, entity_type, db_timestamp);
+```
```
patriknw, line 23: Maybe we should clarify that this migration is optional. If someone is already using it I don't think it would hurt much to just continue with the existing index. Maybe same with the primary key. As mentioned below, there can be a huge performance impact on the db to recreate the index if the data volume is large.

patriknw (commented):
good

docs/src/main/paradox/migration-guide.md
sebastian-alfers: Also, see that this section is currently under the following headline:

> ### Durable State table schema change (optional)

patriknw (approved):
LGTM

PR #562, opened 2024-04-24
fix: insert instead of update new state with updated revision during migration

sebastian-alfers:
Reopening of #554, follow up to https://github.com/akka/akka-persistence-r2dbc/pull/560

Open tasks:
- [x] look into adding tests for additional column and change handler

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
@@ -89,7 +89,7 @@ private[r2dbc] class PostgresDurableStateDao(executorProvider: R2dbcExecutorProv
   import settings.codecSettings.DurableStateImplicits._
   protected def log: Logger = PostgresDurableStateDao.log
 
-  protected val persistenceExt = Persistence(system)
+  private val persistenceExt = Persistence(system)
 
   private val sqlCache = Sql.Cache(settings.numberOfDataPartitions > 1)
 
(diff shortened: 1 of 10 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
```diff
@@ -93,6 +93,8 @@ private[r2dbc] class PostgresDurableStateDao(executorProvider: R2dbcExecutorProv
 
   private val sqlCache = Sql.Cache(settings.numberOfDataPartitions > 1)
 
+  protected def shouldInsert: (SerializedStateRow) => Future[Boolean] = state => Future.successful(state.revision == 1)
```
patriknw, line 96: can't this be an ordinary `def`?

```
protected def shouldInsert(state: SerializedStateRow): Future[Boolean] = {
  Future.successful(state.revision == 1)
}
```

migration-tests/src/test/scala/akka/persistence/r2dbc/migration/MigrationTestColumn.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import akka.persistence.r2dbc.state.scaladsl.AdditionalColumn
+
+class MigrationTestColumn extends AdditionalColumn[Any, String] {
+
+  override val columnName: String = "test_column"
+
+  override def bind(upsert: AdditionalColumn.Upsert[Any]): AdditionalColumn.Binding[String] = {
+    if (upsert.value.asInstanceOf[String] == "s-column") {
+      println(s"??? ${upsert.value.asInstanceOf[String]}")
```
patriknw, line 15: remove?

patriknw (commented):
looking good, just something small...

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
sebastian-alfers: Yes, of course!

patriknw (approved):
LGTM

PR #570, opened 2024-05-15
doc: Sharding vs partitioning

patriknw:
Trigged by some feedback I looked closer at what others say regarding sharding vs partitioning.Seems like sharding is mostly used for horizontal partitioning to distribute the data across multiple servers. While partitioning is a more generic term it mostly refers to splitting up large tables into smaller tables, but still in one database instance (*vertical scaling").

Our primary goal is horizontal scaling so we should highlight database sharding.

That said, the implementation splits the data into multiple tables (partitioning), which can then optionally be located on different database instances (sharding). The documentation still uses data partition and that it corresponds to a table. I think that is fine, but let me know if it's confusing.

### DIFF
--- a/docs/src/main/paradox/index.md
+++ b/docs/src/main/paradox/index.md
@@ -15,7 +15,7 @@ The Akka Persistence R2DBC plugin allows for using SQL database with R2DBC as a 
 * [PostgreSQL JSON](postgres_json.md)
 * [Projection](projection.md)
 * [Configuration](config.md)
-* [Data partitioning](data-partition.md)
+* [Database sharding](data-partition.md)
 * [Cleanup tool](cleanup.md)
 * [Migration tool](migration.md)
 * [Migration Guide](migration-guide.md)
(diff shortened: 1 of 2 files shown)


johanandren (approved):
LGTM, worth backport/republishing the docs to get consistent messaging?

patriknw:
um, was thinking about it, but probably not important enough. Data partitioning is still not wrong, even though end goal is database sharding.

octonato (approved):
LGTM

docs/src/main/paradox/index.md
```diff
@@ -15,7 +15,7 @@ The Akka Persistence R2DBC plugin allows for using SQL database with R2DBC as a
 * [PostgreSQL JSON](postgres_json.md)
 * [Projection](projection.md)
 * [Configuration](config.md)
-* [Data partitioning](data-partition.md)
+* [Database sharding](data-partition.md)
```
pvlugter, line 18: Also change the name of the file for links? Or too much hassle with redirecting?

pvlugter (approved):

docs/src/main/paradox/index.md
patriknw: was also thinking about it, but I don't see it as very important

PR #579, opened 2024-06-07
test: correct sequence number for delete single snapshot test

pvlugter:
Noticed when reusing this test elsewhere: sequence number was incorrect, so it wasn't actually deleting a snapshot. Test still passed as successful if there's no matching snapshot. Rather than just set to the correct sequence number (15) also future proof by checking what it is before delete.

Also fix a typo in the snapshot store docs.


patriknw (approved):
LGTM

Issue #571, opened 2024-05-21
Postgres R2DBC artifact now includes native image metadata


johanandren:
We can probably drop our custom meta and depend on the shipped ones:

```
[error] Deduplicate found different file contents in the following:
[error]   Jar name = akka-persistence-r2dbc_2.13-1.2.4.jar, jar org = com.lightbend.akka, entry target = META-INF/native-image/org.postgresql/r2dbc-postgresql/reflect-config.json
[error]   Jar name = r2dbc-postgresql-1.0.5.RELEASE.jar, jar org = org.postgresql, entry target = META-INF/native-image/org.postgresql/r2dbc-postgresql/reflect-config.json
```

PR #580, opened 2024-06-12
chore: Drop Postgres R2DBC native-image metadata #571

johanandren:
Native image run tested locally ✅


patriknw (approved):
LGTM
Because of some dependency update?

johanandren:
Yeah, I noticed a while ago (#571) that we bumped and they started including metadata but didn't get to it until now.

PR #586, opened 2024-08-01
fix: stuck queries when too many events with same timestamp

pvlugter:
We've seen a case where more than 1000 events were persisted together (so have the same timestamp), and more than the default query buffer size, and then queries will be stuck on the same timestamp and never make further progress.

Since events are ordered by timestamp and sequence number, update the queries to also filter by the highest seen seq number for the latest timestamp (only for events with the same timestamp as this starting timestamp). Fixes the case where all the events are for the same persistence id. Queries will still duplicate on the events with both the same timestamp and seq number (different persistence ids), to handle events across the buffer limit. The very edge case of more events than the buffer size all with the same timestamp and the same seq number (from different persistent ids) would not be handled.

So this always ends up adding an additional filter to queries. We could look at only adding this extra conditional check when the latest timestamp is the same as the previous query (that returned results), to only handle this particular case. The backtracking filtered adjustment (updated in this PR) would need to be aware of this too.

Another alternative is to go further, and always do all the filtering on the database side. Adding conditions for all the latest seen sequence numbers, conditional per persistence id. We would then expect no duplicated events from queries. It complicates the queries even more though, with something like `db_timestamp >= :from_timestamp AND ((persistence_id NOT IN (pid1, pid2, ..., pidN) OR (persistence_id = pid1 AND seq_nr > pid1SeqNr) OR (persistence_id = pid2 AND seq_nr > pid2SeqNr) OR ... OR (persistence_id = pidN AND seq_nr > pidNSeqNr))`. But this would also guarantee progress in all cases, and the additional conditions are only for the 'seen' persistence ids that share the same latest timestamp.

Marking draft while this is under discussion.

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
@@ -74,9 +74,8 @@ import org.slf4j.Logger
     }
   }
 
-  private def highestSeenSeqNr(offset: TimestampOffset): Option[Long] = {
-    if (offset.seen.isEmpty) None else Some(offset.seen.values.max)
-  }
+  private def highestSeenSeqNr(offset: TimestampOffset): Option[Long] =
+    Option.when(offset.seen.nonEmpty)(offset.seen.values.max)
 
   object Buckets {
     type EpochSeconds = Long
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
@@ -31,7 +31,19 @@ import org.slf4j.Logger
 
   object QueryState {
     val empty: QueryState =
-      QueryState(TimestampOffset.Zero, 0, 0, 0, 0, backtrackingCount = 0, TimestampOffset.Zero, 0, 0, Buckets.empty)
+      QueryState(
+        latest = TimestampOffset.Zero,
+        rowCount = 0,
+        rowCountSinceBacktracking = 0,
+        queryCount = 0,
+        idleCount = 0,
+        backtrackingCount = 0,
+        latestBacktracking = TimestampOffset.Zero,
+        latestBacktrackingSeenCount = 0,
+        backtrackingExpectFiltered = 0,
+        buckets = Buckets.empty,
+        previous = TimestampOffset.Zero,
+        previousBacktracking = TimestampOffset.Zero)
   }
 
   final case class QueryState(
@@ -44,7 +56,9 @@ import org.slf4j.Logger
       latestBacktracking: TimestampOffset,
       latestBacktrackingSeenCount: Int,
       backtrackingExpectFiltered: Int,
-      buckets: Buckets) {
+      buckets: Buckets,
+      previous: TimestampOffset,
+      previousBacktracking: TimestampOffset) {
 
     def backtracking: Boolean = backtrackingCount > 0
 
@@ -57,8 +71,8 @@ import org.slf4j.Logger
       else latest.timestamp
 
     def nextQueryFromSeqNr: Option[Long] =
-      if (backtracking) highestSeenSeqNr(latestBacktracking)
-      else highestSeenSeqNr(latest)
+      if (backtracking) highestSeenSeqNr(previousBacktracking, latestBacktracking)
+      else highestSeenSeqNr(previous, latest)
 
     def nextQueryToTimestamp(atLeastNumberOfEvents: Int): Option[Instant] = {
       buckets.findTimeForLimit(nextQueryFromTimestamp, atLeastNumberOfEvents) match {
@@ -74,8 +88,11 @@ import org.slf4j.Logger
     }
   }
 
-  private def highestSeenSeqNr(offset: TimestampOffset): Option[Long] =
-    Option.when(offset.seen.nonEmpty)(offset.seen.values.max)
+  // only filter by highest seen seq nr when the next query is the same timestamp (or when unknown for initial queries)
+  private def highestSeenSeqNr(previous: TimestampOffset, latest: TimestampOffset): Option[Long] =
+    Option.when((previous == TimestampOffset.Zero || previous.timestamp == latest.timestamp) && latest.seen.nonEmpty) {
+      latest.seen.values.max
+    }
 
   object Buckets {
     type EpochSeconds = Long
@@ -220,10 +237,10 @@ import org.slf4j.Logger
       // so continue until rowCount is 0. That means an extra query at the end to make sure there are no
       // more to fetch.
       if (state.queryCount == 0L || state.rowCount > 0) {
-        val newState = state.copy(rowCount = 0, queryCount = state.queryCount + 1)
+        val newState = state.copy(rowCount = 0, queryCount = state.queryCount + 1, previous = state.latest)
 
         val fromTimestamp = state.latest.timestamp
-        val fromSeqNr = highestSeenSeqNr(state.latest)
+        val fromSeqNr = highestSeenSeqNr(state.previous, state.latest)
 
         val toTimestamp = newState.nextQueryToTimestamp(settings.querySettings.bufferSize) match {
           case Some(t) =>
@@ -325,7 +342,8 @@ import org.slf4j.Logger
 
         val newSeenCount =
           if (offset.timestamp == state.latestBacktracking.timestamp &&
-            highestSeenSeqNr(offset) == highestSeenSeqNr(state.latestBacktracking))
+            highestSeenSeqNr(state.previousBacktracking, offset) ==
+              highestSeenSeqNr(state.previousBacktracking, state.latestBacktracking))
             state.latestBacktrackingSeenCount + 1
           else 1
 
@@ -462,7 +480,11 @@ import org.slf4j.Logger
           else s"Found [${state.rowCount}] rows in previous query.")
       }
 
-      newState ->
+      val newStateWithPrevious =
+        if (newState.backtracking) newState.copy(previousBacktracking = newState.latestBacktracking)
+        else newState.copy(previous = newState.latest)
+
+      newStateWithPrevious ->
       Some(
         dao
           .rowsBySlices(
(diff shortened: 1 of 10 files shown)


pvlugter:
SQL Server tests not able to start the database to begin with. Looks like this path `/opt/mssql-tools/bin/sqlcmd` is now `/opt/mssql-tools18/bin/sqlcmd`. Updating that but then it fails with:

```
Sqlcmd: Error: Microsoft ODBC Driver 18 for SQL Server : TCP Provider: Error code 0x2749.
Sqlcmd: Error: Microsoft ODBC Driver 18 for SQL Server : SSL Provider: [error:0A000086:SSL routines::certificate verify failed:self-signed certificate].
Sqlcmd: Error: Microsoft ODBC Driver 18 for SQL Server : A network-related or instance-specific error has occurred while establishing a connection to localhost. Server is not found or not accessible. Check if instance name is correct and if SQL Server is configured to allow remote connections. For more information see SQL Server Books Online..
```

Will try pinning to an earlier version instead.

core/src/test/scala/akka/persistence/r2dbc/query/EventsBySliceSpec.scala
```diff
@@ -168,6 +214,64 @@ class EventsBySliceSpec
...
+            .runWith(sinkProbe)
+            .request(10)
+
+        def take(n: Int): Set[String] =
+          (1 to n).map(_ => result.expectNext().event).toSet
+
+        take(2) shouldBe Set("A1", "A2")
+        take(3) shouldBe Set("A3", "B3", "C3")
+        take(2) shouldBe Set("A4", "B4")
+        take(2) shouldBe Set("A5", "A6")
```
pvlugter, line 270: Postgres does return the events with the same seq number in the inserted order, but probably shouldn't rely on that for tests.

octonato (approved):
LGTM

...with a few suggestions

johanandren (commented):
Nice!

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -70,6 +74,10 @@ import org.slf4j.Logger
     }
   }
 
+  private def highestSeenSeqNr(offset: TimestampOffset): Option[Long] = {
+    if (offset.seen.isEmpty) None else Some(offset.seen.values.max)
+  }
```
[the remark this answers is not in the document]
pvlugter: Sure. Still have a habit of being scala 2.12 compatible.

core/src/main/scala/akka/persistence/r2dbc/internal/h2/H2QueryDao.scala
```diff
@@ -24,13 +24,17 @@ private[r2dbc] class H2QueryDao(executorProvider: R2dbcExecutorProvider) extends
...
+      fromSeqNrParam: Boolean,
       toDbTimestampParam: Boolean,
       behindCurrentTime: FiniteDuration,
       backtracking: Boolean,
       minSlice: Int,
       maxSlice: Int): String = {
     // not caching, too many combinations
 
+    def fromSeqNrParamCondition =
+      if (fromSeqNrParam) "AND (db_timestamp != ? OR seq_nr >= ?)" else ""
```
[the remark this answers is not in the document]
pvlugter: Ok. Will add a comment about why the seq number check is only for the same timestamp. And I guess the conditional is not as obvious in this logical form, compared to say `if` in code. This can be read along the lines of `if (db_timestamp == from_timestamp) seq_nr >= from_seq_nr`, with `P -> Q === ~P v Q`

core/src/test/scala/akka/persistence/r2dbc/query/EventsBySliceSpec.scala
```diff
@@ -168,6 +214,64 @@ class EventsBySliceSpec
...
+        val timestamp = InstantFactory.now()
+        writeEvent(slice1, pid1, 1L, timestamp, "A1")
+        writeEvent(slice1, pid1, 2L, timestamp, "A2")
+        writeEvent(slice1, pid1, 3L, timestamp, "A3")
+        writeEvent(slice1, pid1, 4L, timestamp, "A4")
+        writeEvent(slice1, pid1, 5L, timestamp, "A5")
+        writeEvent(slice1, pid1, 6L, timestamp, "A6")
+        writeEvent(slice2, pid2, 3L, timestamp, "B3")
+        writeEvent(slice2, pid2, 4L, timestamp, "B4")
+        writeEvent(slice3, pid3, 3L, timestamp, "C3")
```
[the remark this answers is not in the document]
pvlugter: It is covered by existing tests, but useful to cover explicitly here. Will add this.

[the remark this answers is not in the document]
pvlugter: Extended test with D1, D2, D3 on later timestamp.

core/src/main/scala/akka/persistence/r2dbc/internal/h2/H2QueryDao.scala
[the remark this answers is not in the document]
pvlugter: Added an explanatory comment. Could also be expressed with `NOT (db_timestamp = ? AND seq_nr < ?)` — not sure if that's more obvious though. Comment notes that seq numbers are per persistence id.

johanandren (approved):

pvlugter:
Test failing with `docker-compose: command not found`. Underlying image has been updated by GitHub, removing the old `docker-compose` command? Will update to `docker compose`.

pvlugter:
I've updated to track the timestamp from the previous query, and only apply the extra seq nr filter when it's already known to be the same timestamp for the next query: https://github.com/akka/akka-persistence-r2dbc/pull/586/commits/09e05fcc8b3c1b90aa18525b1990c9c3882da9ab

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -70,6 +88,12 @@ import org.slf4j.Logger
     }
   }
 
+  // only filter by highest seen seq nr when the next query is the same timestamp (or when unknown for initial queries)
+  private def highestSeenSeqNr(previous: TimestampOffset, latest: TimestampOffset): Option[Long] =
+    Option.when((previous == TimestampOffset.Zero || previous.timestamp == latest.timestamp) && latest.seen.nonEmpty) {
+      latest.seen.values.max
```
patriknw, line 94: > Note that sequence numbers are per persistence id, so a later timestamp can have an earlier sequence number

That is correct. How is that handled? I guess it is caught by backtracking in the same way when visibility of pid-A may be seen before another pid-B even though timestamp of pid-B is before pid-A.

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -446,14 +480,19 @@ import org.slf4j.Logger
...
         dao
           .rowsBySlices(
             entityType,
             minSlice,
             maxSlice,
             fromTimestamp,
+            fromSeqNr,
             toTimestamp,
             behindCurrentTime,
             backtracking = newState.backtracking)
```
patriknw, line 498: Do we need to adjust `BySliceQuery.deserializeAndAddOffset`? It has a check on the buffer size and throws IllegalStateException

patriknw (commented):
looking good

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
pvlugter: If the events are there, it handles it by only filtering by highest sequence number for the same timestamp as the query starting timestamp. See also where the query is adjusted, and the comment there.

So if there are events (ordered by timestamp, seq nr):

timestamp | seq nr | pid  | event
----------|--------|------|------
t1        | 1      | pid1 | A1
t1        | 1      | pid2 | B1
t1        | 1      | pid3 | C1
t1        | 2      | pid1 | A2
t1        | 2      | pid2 | B2
t1        | 3      | pid1 | A3
t2        | 2      | pid3 | C2
t2        | 3      | pid2 | B3
t2        | 4      | pid1 | A4

And a buffer size of 4. First query only gets to the 4th event (A2). Before it would restart from `timestamp >= t1`, stuck on repeating the same 4 events (A1, B1, C1, A2). Now it will use the query `timestamp >= t1 AND (timestamp != t1 OR seq_nr >= 2)` for the next query, starting from the 4th event now, processing (A2 (deduplicated), B2, A3, C2). The seq number filter only applies to the starting timestamp, so that the query is otherwise just by timestamp as usual.

Applies for the both the regular queries and the backtracking queries. Backtracking catching any late arriving events as usual.

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
pvlugter: That check is whether the seen map exceeds the buffer size (more persistence ids on the same timestamp than the buffer size). We could adjust it to only throw if all the sequence numbers are the same, which would not be handled by this fix (would be stuck on both timestamp and seq number). But that many persistence ids with events on the exact same timestamp already feels exceptional, so thought that it's useful to leave as is.

pvlugter:
Also discussed on internal issue: we may want to have a limit on the number of events that can be persisted together in the first place, as these are committed atomically in a transaction, and it could indicate a bug in user code.

Or if we want to support writing large numbers of events in the same operation, we could group events in batches with separate transactions (wouldn't be atomic across all events though).

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
patriknw: Good, that should work fine. What I was thinking about would be classified as late arriving event.

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
patriknw: I agree 👍

patriknw (approved):
LGTM

PR #599, opened 2024-09-18
feat: Accept PubSub events after idle

patriknw:
* PubSub events are ignored if they are too far ahead of backtracking.
* That means that they will always be ignored after an idle period.
* This emits heartbeat events when the query is idle and thereby progress the backtracking timestamp in the PubSub filter.

This is just a sketch so far. I haven't tried it at all yet.


core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -149,6 +154,22 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+      TimestampOffset(timestamp, Map.empty),
+      "", // FIXME persistenceId
+      1L,
+      eventOption = Some(event),
+      timestamp.toEpochMilli,
+      eventMetadata = None,
+      "", // FIXME entityType
+      0, // FIXME slice
+      filtered = true,
+      source = EnvelopeOrigin.SourceHeartbeat,
```
patriknw, line 169: probably more places downstreams where we need to handle this new source type

core/src/main/scala/akka/persistence/r2dbc/internal/ContinuousQuery.scala
```diff
@@ -150,8 +153,12 @@ final private[r2dbc] class ContinuousQuery[S, T](
                   next()
                 }
             })
+            val sourceWithHeartbeat = heartbeat(newState) match {
+              case None    => source
+              case Some(h) => Source.single(h).concat(source)
```
patriknw, line 158: maybe a little strange to emit it before a real query, but I didn't see an easy way to do it after a query with a fully updated state

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -502,12 +503,22 @@ import org.slf4j.Logger
           .via(deserializeAndAddOffset(newState.currentOffset)))
     }
 
+    def heeartbeat(state: QueryState): Option[Envelope] = {
+      if (state.idleCount >= 1) {
+        val timestamp = state.latestBacktracking.timestamp.plusMillis(
+          settings.querySettings.refreshInterval.toMillis * state.idleCount)
```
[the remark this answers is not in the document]
patriknw: better to keep track of wall clock since the latestBacktracking

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -502,12 +503,22 @@ import org.slf4j.Logger
           .via(deserializeAndAddOffset(newState.currentOffset)))
     }
 
+    def heeartbeat(state: QueryState): Option[Envelope] = {
```
pvlugter, line 506:
```suggestion
    def heartbeat(state: QueryState): Option[Envelope] = {
```

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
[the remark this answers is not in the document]
pvlugter: sounds good to just keep track of wall clock time since

pvlugter (commented):
👍🏼 direction looks good to me

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -110,19 +112,52 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+        val pid = generateHeartbeatPersistenceId(entityType, slice)
+        heartbeatPersistenceIds.put(key, pid)
+        pid
+      case pid => pid
+    }
+  }
+
+  @tailrec private def generateHeartbeatPersistenceId(entityType: String, slice: Int, n: Int = 1): String = {
+    if (n < 1000000) {
+      val pid = PersistenceId.concat(entityType, s"_ht-$n")
```
pvlugter, line 132: Should we use `_hb` here, to align with the `HB` for the source name?

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -110,19 +112,52 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+      if (persistenceExt.sliceForPersistenceId(pid) == slice)
+        pid
+      else
+        generateHeartbeatPersistenceId(entityType, slice, n + 1)
+    } else
+      throw new IllegalStateException(s"Couldn't find heartbeat persistenceId for [$entityType] with slice [$slice]")
+
+  }
```
pvlugter, line 140: This looks like the most awkward part of the solution. Needing to find a dummy persistence id that maps to the slice.

Could we have a completely custom persistence id instead, that doesn't include the entityType? So that the persistence id still hashes to the min slice correctly, and the entityType is in the envelope already.  Then it could be generated upfront as a constant. Or do we rely on extracting the entity type from the persistence id in too many places?

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -502,12 +519,23 @@ import org.slf4j.Logger
           .via(deserializeAndAddOffset(newState.currentOffset)))
     }
 
+    def heartbeat(state: QueryState): Option[Envelope] = {
+      if (state.idleCount >= 1 && state.latestBacktrackingWallClock != Instant.EPOCH) {
+        // using wall clock to measure duration since the last backtracking event
+        val timestamp =
+          state.latestBacktracking.timestamp.plus(JDuration.between(state.latestBacktrackingWallClock, clock.instant()))
+        createHeartbeat(timestamp)
```
pvlugter, line 527: Ok, so rather than just wall clock time, we make this relative to what could be the db timestamp. Makes sense.

pvlugter (approved):
LGTM. Don't know if we can improve that persistence id generation.

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
pvlugter: Probably not frequent enough to worry about though.

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
patriknw: sure, do you think it is unique enough to not have collision with a user's pid? Otherwise we could make it longer by adding a uuid, either a fixed one or a new one per R2dbcReadJournal instance (same lifecycle as the cache)

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
patriknw: I'm worried about that the entityType extraction (and also the slice from pid) is used in many places, since everything is scoped around the entityType, such as the query itself and sharding. Might work but high risk.

Since it's cached with the same lifecycle as R2dbcReadJournal instance (typically one per ActorSystem) it will not be very many.

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -110,19 +113,55 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+        heartbeatPersistenceIds.put(key, pid)
+        pid
+      case pid => pid
+    }
+  }
+
+  @tailrec private def generateHeartbeatPersistenceId(entityType: String, slice: Int, n: Int = 1): String = {
+    if (n < 1000000) {
+      // including a uuid to make sure it is not the same as any persistence id of the application
+      val pid = PersistenceId.concat(entityType, s"_hb-$heartbeatUuid-$n")
```
patriknw, line 136: I added a uuid in the pid. Shouldn't be high frequency of these. Now it's one per idle tick (per 3 seconds), but we could emit less often. Could be nice to be able to track correlation via an uuid as well. Not a big deal, could change to `s"_hb-akka-$n"`

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
pvlugter: There are some tricks we could do (around how string hashcode is calculated) to avoid needing to brute force it. But that's something we could come back to later on. No problem to change what the heartbeat persistence id is, providing it still maps to the right slice.

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
pvlugter: Seems ok. I guess we could also make it more unique with its format and mark as akka internal. But can be revisited at any time.

pvlugter:
> having a hard time understanding potential bad consequences

Yeah. I think needing to extend it within the existing constraints, with dummy events and special persistence id, rather than being able to have a different envelope type or something, makes it more difficult.

pvlugter (approved):

patriknw:
This should probably not go in to a patch release since there could be cases in Akka Projection where it's not filtered out completely. https://github.com/akka/akka-projection/pull/1200 fixes that and we can include this and that in next minor versions

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
pvlugter: So this can create some quite old timestamps. If replaying older events, then the latest backtracking timestamp can be from some time ago, while latest backtracking wall clock is from when that offset was seen. So whatever the distance is between those, we end up that far back from the current time.

pvlugter: So I assume we wanted to avoid using the current wall clock instant, in case it's skewed from the database timestamps for events. But if we have a big enough gap between the event timestamp and when it was seen by backtracking, then the heartbeats will be outside the backtracking window and published events will need to be filtered. Won't resolve itself until new backtracking envelopes are seen, as before.

pvlugter: Could be tricky to have a relative timestamp for all cases, as it needs to be from an event that was not originally written too long ago. With backtracking events, only works when it's up to date, and these are seen close enough to their write times. To have a more accurate marker for the latest db timestamp, maybe this could use published events instead — they'll be arriving soon after their writes. If we use the last seen published event + wall clock difference, then it could give an accurate relative heartbeat time.

pvlugter: I guess basing it all in published events would need to be quite different. Mostly tracking it together in the pubsub event filtering, with just an empty heartbeat on idle coming from the query here. And maybe opens up the possibility of heartbeats actually getting ahead of later published events?

Maybe this should be quite simple. Get the current timestamp from the database when the query starts, and then just create relative timestamps from that start timestamp for each heartbeat on idle.

pvlugter: Switched to using an initial database timestamp in #612.

patriknw:

PR #600, opened 2024-09-20
feat: Skip backtracking when far behind

patriknw:


core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -416,20 +416,28 @@ import org.slf4j.Logger
...
+      // Note that when starting the query with offset = NoOffset it will switch to backtracking immediately after
+      // the first normal query because between(latestBacktracking.timestamp, latest.timestamp) > halfBacktrackingWindow
+
+      val qSettings = settings.querySettings
+      import akka.util.JavaDurationConverters._
+
+      qSettings.backtrackingEnabled &&
+      !state.backtracking &&
+      state.latest != TimestampOffset.Zero &&
+      state.latest.timestamp.isAfter(clock.instant().minus(qSettings.backtrackingWindow.asJava)) &&
```
patriknw, line 429: Have to think about the exact timestamps for this. One candidate is to look at the `toTimestamp`, but as the code is structured now that is kind calculated afterwards.

core/src/test/scala/akka/persistence/r2dbc/query/EventsBySliceBacktrackingSpec.scala
```diff
@@ -246,6 +249,63 @@ class EventsBySliceBacktrackingSpec
...
+
+      expect(result1.expectNext(), pid1, 101, Some("e1-101"))
+      expect(result1.expectNext(), pid2, 101, Some("e2-101"))
+
+      // backtracking events
+      expect(result1.expectNext(), pid1, 101, None)
+      expect(result1.expectNext(), pid2, 101, None)
+
+      result1.cancel()
+    }
```
patriknw, line 307: this test is passing, but some others are failing now

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
patriknw: must also adjust the `latestBacktracking` because otherwise when backtracking is activated again it would start from the far behind latestBacktracking

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -416,20 +420,30 @@ import org.slf4j.Logger
...
+
+      val qSettings = settings.querySettings
+      val previousTimestamp =
+        if (state.previous == TimestampOffset.Zero) state.latest.timestamp else state.previous.timestamp
+
+      qSettings.backtrackingEnabled &&
+      !state.backtracking &&
+      state.latest != TimestampOffset.Zero &&
+      // no backtracking if far behind current wall clock time
+      previousTimestamp.isAfter(clock.instant().minus(firstBacktrackingQueryWindow)) &&
```
patriknw, line 435: I don't think it has to be more exact than this.

I have also verified that this works as expected with a sample app. Generated some 100k events test data. Ran catch up without triggering any backtracking, and then storing some new events that trigger backtracking again, and after a while backtracking stops again.

patriknw: No, there is one problem. Let's say that a projection is stopped/crashed and then isn't started immediately again. Then it should still run backtracking from the start offset (minus some) until the backtracking offset is ahead of the start offset.

patriknw:

Issue #546, opened 2024-04-05
failed: EventsBySlicePubSubSpec: should turn FilteredPayload to filtered events with no payload


ennru:
https://github.com/akka/akka-persistence-r2dbc/actions/runs/8565523378/job/23473696096#step:7:1209

```
[info] - should turn FilteredPayload to filtered events with no payload *** FAILED *** (100 milliseconds)
[info]   java.lang.AssertionError: expected OnNext, found OnError(java.lang.IllegalArgumentException: Slice range [0-1023] spans over more than one of the [4] data partitions.
[info] 	at akka.persistence.r2dbc.internal.postgres.PostgresQueryDao.rowsBySlices(PostgresQueryDao.scala:227)
[info] 	at akka.persistence.r2dbc.internal.BySliceQuery.nextQuery$2(BySliceQuery.scala:459)
```

johanandren:
https://github.com/akka/akka-persistence-r2dbc/actions/runs/11127532266/job/30920059443#step:7:1226

PR #604, opened 2024-10-02
test: FilteredPayload with more db partitions, #546

patriknw:
Refs #546


core/src/test/scala/akka/persistence/r2dbc/query/EventsBySlicePubSubSpec.scala
```diff
@@ -200,6 +200,8 @@ class EventsBySlicePubSubSpec
     }
 
     "turn FilteredPayload to filtered events with no payload" in new Setup {
+      pendingIfMoreThanOneDataPartition()
```
patriknw, line 203: slice range 0-1023 can't be used with more than one db partition

johanandren (approved):

PR #612, opened 2024-10-16
fix: base relative heartbeat timestamps on initial database time

pvlugter:


core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -539,22 +540,33 @@ import org.slf4j.Logger
     }
 
     def heartbeat(state: QueryState): Option[Envelope] = {
-      if (state.idleCount >= 1 && state.latestBacktrackingWallClock != Instant.EPOCH) {
-        // using wall clock to measure duration since the last backtracking event
-        val timestamp =
-          state.latestBacktracking.timestamp.plus(JDuration.between(state.latestBacktrackingWallClock, clock.instant()))
+      if (state.idleCount >= 1 && state.latestBacktracking != TimestampOffset.Zero) {
+        // using wall clock to measure duration since the start time (database timestamp)
+        val timestamp = state.startTime.plus(JDuration.between(state.startTimeWallClock, clock.instant()))
```
pvlugter, line 545: Maybe also subtract the backtracking behind current time. And could be when the last query actually ran, rather than now instant.

patriknw: sounds good

pvlugter: Was going to base this from the previous backtracking query wall clock minus backtracking behind current time — so that heartbeat is the guaranteed edge of backtracking — but that doesn't work when backtracking is disabled when it falls behind. So instead using previous wall clock for any query, so that it's the limit of where backtracking would be if that query was for backtracking. And still subtracting backtracking behind current time, so that heartbeats (and pubsub events) can't get too far ahead. I think that makes sense, but interactions are getting more complicated.

patriknw (approved):
LGTM

PR #613, opened 2024-10-18
fix: filter heartbeats early in pubsub merge

pvlugter:
Filter heartbeats once received by the skip pubsub if too far ahead check.

Remove some extra filtering from tests again, which should fail if heartbeats are emitted.

Have left the heartbeat creation as is, including the search for appropriate persistence id, but heartbeats could also be simpler dummy envelopes now that they're not going further and to the offset store.


patriknw (approved):
LGTM

patriknw:

PR #615, opened 2024-10-18
feat: Skip backtracking, PubSub after idle (backport 1.2)

patriknw:
Backport of #614

* feat: Accept PubSub events after idle
  * PubSub events are ignored if they are too far ahead of backtracking.
  * That means that they will always be ignored after an idle period.
  * This emits heartbeat events when the query is idle and thereby progress the backtracking timestamp in the PubSub filter.
  * base relative heartbeat timestamps on initial database time
  * previous query wall clock minus backtracking behind current time
  * including fake persistenceId corresponding to the slice
  * don't deduplicate heartbeat
  * mima filter
  * unique heartbeat persistenceId
  * filter heartbeats early in pubsub merge

* feat: Skip backtracking when far behind

---------

Co-authored-by: Peter Vlugter <59895+pvlugter@users.noreply.github.com>
(cherry picked from commit c7899d1318d69b872b918aea117aabfd7e300c43)


patriknw:
hm, why isn't ci running here

PR #630, opened 2024-11-29
fix: don't trigger heartbeats from idle backtracking

pvlugter:
Port of https://github.com/akka/akka-persistence-dynamodb/pull/104


patriknw (approved):
LGTM

PR #636, opened 2025-01-13
chore: remove obsolete outside backtracking debug log

pvlugter:
Now that we disable backtracking when far behind, this old debug log is not so useful — will log every offset in catch up mode. Removing it. Alternative would be to explicitly track when in catch up mode with disabled backtracking in the state, so we only log this when backtracking is enabled.


patriknw (approved):
LGTM

octonato (approved):

Issue #637, opened 2025-01-14
Backtracking start timestamp when enabled again


patriknw:
Not sure if this is a real issue yet, but came to think of it. What is the start timestamp for the backtrack query when we disable backtracking in catchup mode and then enable it again? Is there a case where we run backtracking (state.latestBacktracking is defined), then disable backtracking and then enable it again?

pvlugter:
Yes, the enable-disable-enable always happens when far behind. There's always initial backtracking, so it will have backtracking enabled (and latest backtracking is defined), then disable, then enable again when caught up.

I remember checking this when testing the disabled backtracking. A check was added specifically for this case, and it will limit to latest minus backtracking window:

https://github.com/akka/akka-persistence-r2dbc/blob/25d3220751a58181ca0017f0bf6ae691dee50653/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala#L82-L84

PR #642, opened 2025-01-14
test: Small addition to disabled backtracking test

patriknw:
Refs https://github.com/akka/akka-persistence-r2dbc/issues/637

Seems to work as expected.

### DIFF
--- a/core/src/test/scala/akka/persistence/r2dbc/query/EventsBySliceBacktrackingSpec.scala
+++ b/core/src/test/scala/akka/persistence/r2dbc/query/EventsBySliceBacktrackingSpec.scala
@@ -359,6 +359,17 @@ class EventsBySliceBacktrackingSpec
       // no backtracking
       result1.expectNoMessage()
 
+      val now = InstantFactory.now().minus(settings.querySettings.backtrackingBehindCurrentTime.toJava)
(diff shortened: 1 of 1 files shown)


core/src/test/scala/akka/persistence/r2dbc/query/EventsBySliceBacktrackingSpec.scala
```diff
@@ -359,6 +359,17 @@ class EventsBySliceBacktrackingSpec
...
+      writeEvent(slice1, pid1, 31, now.plusMillis(1), "e1-31")
+      writeEvent(slice2, pid2, 31, now.plusMillis(2), "e2-31")
+
+      expect(result1.expectNext(), pid1, 31, Some("e1-31"))
+      expect(result1.expectNext(), pid2, 31, Some("e2-31"))
+
+      // backtracking events
+      expect(result1.expectNext(), pid1, 31, None)
+      expect(result1.expectNext(), pid2, 31, None)
+
```
patriknw, line 372: this is actually already tested in the other "skip backtracking when far behind current time", with the difference of using NoOffset as start offset

pvlugter (approved):

PR #645, opened 2025-01-15
fix: Progress bucket count query if empty

patriknw:
* the bucket queries continues from the end of the previous
* the limit of 10000 times 10 seconds is around 28 hours
* meaning that if there were no new events for 28 hours the queries will not progress since it will not find any more
* solution here is to append empty bucket at the end in that case

Follow up on https://github.com/akka/akka-persistence-r2dbc/pull/641

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
@@ -222,6 +222,21 @@ import org.slf4j.Logger
         fromTimestamp: Instant,
         limit: Int): Future[Seq[Bucket]]
 
+    protected def appendTwoEmptyBucketsIfMissing(
+        buckets: IndexedSeq[Bucket],
+        toTimestamp: Instant): IndexedSeq[Bucket] = {
+      import Buckets.BucketDurationSeconds
+      val startTimeOfLastBucket = (toTimestamp.getEpochSecond / 10) * 10
+      val startTimeOf2ndLastBucket = ((toTimestamp.getEpochSecond - BucketDurationSeconds) / 10) * 10
+      if (buckets.last.startTime != startTimeOfLastBucket && buckets.reverseIterator
+          .drop(1)
+          .next()
+          .startTime != startTimeOf2ndLastBucket)
+        buckets :+ Bucket(startTimeOf2ndLastBucket, 0) :+ Bucket(startTimeOfLastBucket, 0)
+      else
+        buckets
+    }
+
   }
 }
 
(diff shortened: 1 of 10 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -222,6 +222,21 @@ import org.slf4j.Logger
         fromTimestamp: Instant,
         limit: Int): Future[Seq[Bucket]]
 
+    protected def appendTwoEmptyBucketsIfMissing(
+        buckets: IndexedSeq[Bucket],
+        toTimestamp: Instant): IndexedSeq[Bucket] = {
+      import Buckets.BucketDurationSeconds
+      val startTimeOfLastBucket = (toTimestamp.getEpochSecond / 10) * 10
+      val startTimeOf2ndLastBucket = ((toTimestamp.getEpochSecond - BucketDurationSeconds) / 10) * 10
```
patriknw, line 230: There could be more cases that are not covered by this, such as that last bucket is found but not 2nd last.

nextStartTime that was added in https://github.com/akka/akka-persistence-r2dbc/pull/641 is using the 2nd last, but it can be simplified by using the last - BucketDurationSeconds instead. Then I can just add the last bucket here if that is missing.

core/src/test/scala/akka/persistence/r2dbc/TestDbLifecycle.scala
```diff
@@ -74,4 +77,35 @@ trait TestDbLifecycle extends BeforeAndAfterAll { this: Suite =>
     super.beforeAll()
   }
 
+  // to be able to store events with specific timestamps
+  def writeEvent(slice: Int, persistenceId: String, seqNr: Long, timestamp: Instant, event: String): Unit = {
```
patriknw, line 81: moved this here since it was used in several tests

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
patriknw: https://github.com/akka/akka-persistence-r2dbc/pull/645/commits/77a6e0c39fcb72dedabee783ca65307f7a8b6bcd

core/src/test/scala/akka/persistence/r2dbc/query/BucketCountSpec.scala
```diff
@@ -81,14 +81,12 @@ class BucketCountSpec
           .countBuckets(entityType, 0, persistenceExt.numberOfSlices - 1, startTime, limit)
           .futureValue
       withClue(s"startTime $startTime ($bucketStartTime): ") {
-        buckets.size shouldBe 12
+        buckets.size shouldBe 1
```
pvlugter, line 84: Should be 11?

pvlugter:
```suggestion
        buckets.size shouldBe 11
```

pvlugter (approved):
LGTM

pvlugter:
Also needs a mima filter.

PR #652, opened 2025-01-27
Revert "bump: scala-library 2.13.16 (was 2.13.15) (#648)"

patriknw:
This reverts commit 080467538248db24823d3d5745c5f2e976422717.

Reverts https://github.com/akka/akka-persistence-r2dbc/pull/648 for now because it enforces updates to all downstream usage, including end users. We'll probably have to bump Scala versions in a more coordinated way, from the ground up.

```
scalaInstance) expected `scalaVersion` to be "2.13.16" or later,
[error] but found "2.13.15"; upgrade scalaVersion to fix the build.
[error]
[error] to support backwards-only binary compatibility (SIP-51),
[error] the Scala 2.13 compiler cannot be older than scala-library on the
[error] dependency classpath.
```

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -9,6 +9,7 @@ updates.ignore = [
   { groupId = "org.scalameta", artifactId = "scalafmt-core" },
   { groupId = "org.scalameta", artifactId = "sbt-scalafmt" }
   { groupId = "com.typesafe.akka" },
+  { groupId = "org.scala-lang" }
 ]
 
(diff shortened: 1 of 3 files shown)


pvlugter (approved):
Could be useful to start adding to scala steward ignores in all projects, if we'll always need to coordinate manual bumps.

patriknw:
ok, added ignore. Given this error we will notice in downstream projects when we bump in Akka core.

PR #660, opened 2025-02-12
feat: loading last event in R2dbcReadJournal

aludwiko:


core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -911,6 +911,23 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
...
+   *   The persistence id to load the last event for.
+   * @param toSeqNr
+   *   The sequence number to load the last event up to.
+   * @param includeDeleted
+   *   Whether to include deleted events.
+   */
+  def loadLastEvent(
+      persistenceId: String,
+      toSeqNr: Long,
+      includeDeleted: Boolean): Future[Option[ClassicEventEnvelope]] = {
```
patriknw, line 927: Let's use the typed.EventEnvelope instead, see currentEventsByPersistenceIdTyped and use deserializeBySliceRow.

`includeDeleted` is a rather internal thing that shouldn't be needed in public api. We can always use false for this.

patriknw (approved):
LGTM

PR #667, opened 2025-04-11
feat: support latest event timestamp queries for observability

pvlugter:


pvlugter:
Only tested with Postgres locally. May need some adjustments for other dialects.

patriknw (commented):
looking good, sqlserver requires a different `deleted = $sqlFalse`

patriknw:
Added the query interface in Akka core https://github.com/akka/akka/pull/32700
We need that because of gRPC projections https://github.com/akka/akka-projection/pull/1334

Pushed implementation of that interface to this PR

pvlugter:
Updated to published akka snapshot (so tests can run) and fixed for sqlserver. Other than an actual akka release, this should be ready now.

patriknw (approved):
LGTM, I have updated to Akka core 2.10.4

PR #670, opened 2025-04-17
perf: Use select per slice for latestEventTimestamp

patriknw:
* otherwise the planner is reverting to seq scans

Follow up to https://github.com/akka/akka-persistence-r2dbc/pull/667

Credit to @pvlugter for analyzing the query and suggesting the union all solution.


octonato:
I merged #671 (alternative to this), see also [comment in #3701](https://github.com/lightbend/kalix-runtime/issues/3701#issuecomment-2816445189).

Therefore, closing this one.

PR #690, opened 2025-07-06
Add a warning for JSONB and Remember Entities

JustinPihony:
Because Remember Entities don't use JSON for serialization then this causes conflicts


pvlugter (approved):
LGTM, with changed to `extref`.

octonato:
I applied the suggested changes. Will merge when green.

octonato (approved):

PR #705, opened 2025-11-19
fix: ensure queries switch from backtracking

pvlugter:
We've seen a particular case where queries are stuck cycling in backtracking without making progress. Conditions needed to trigger this:

- events persisted in batch with the same timestamp, with size of batch **exactly equal to** query buffer size
- and regular query ends on that same timestamp and switches to backtracking
- and backtracking query start is reset (after being idle for more than backtracking window or on projection start)

Test reproduces scenario by persisting exact events on projection start, with small buffer size. But can also be seen during regular processing if this combination comes together.

The issue is in the switch-from-backtracking check, where it considers backtracking complete when the row count returned from a query is less than the max expected events (buffer size - expect filtered). In the problematic scenario, the backtracking query sees all of the events with the same timestamp in one set of results, marking them all as expected to be filtered on next query. There is already handling for events in a batch >= buffer size, where it switches to starting from both timestamp and seq number (for same timestamp), and it adjusts the expected filtered events for this case. However, this adjustment to expect filtered is applied when processing offsets, and in this case all events have been filtered out. So while the backtracking query has adjusted for this, it's both bounded by the regular query latest on same timestamp (so it doesn't see any new events) and the switch-from-backtracking check still expects all events to be filtered.

Fix this in two ways: first, always use the from-seq-number queries (when available on the latest offset). Previously it was looking to be efficient by only using these queries after having a buffer full of events with the same timestamp, avoiding the extra condition on regular queries, but needing to make an additional query with the adjustment and having more duplicate events when these are batched. This also simplifies the expect filtered calculation, to always be based on timestamp and seq nr. The expect filtered will nearly always be 1 now. Second fix is to guarantee that we switch-from-backtracking when result row count (after filtering) is 0, as an additional guard against this situation.

### DIFF
--- a/core/src/test/scala/akka/persistence/r2dbc/query/EventsBySliceBacktrackingSpec.scala
+++ b/core/src/test/scala/akka/persistence/r2dbc/query/EventsBySliceBacktrackingSpec.scala
@@ -415,31 +415,98 @@ class EventsBySliceBacktrackingSpec
           .request(BufferSize * 2 + 1)
 
       // process all events with normal query
-      val lastOffset = (1 to BufferSize).foldLeft(NoOffset: Offset) { (_, seqNr) =>
+      val lastOffset1 = (1 to BufferSize).foldLeft(TimestampOffset.Zero) { (_, seqNr) =>
         val envelope = result.expectNext()
         envelope.persistenceId shouldBe pid
         envelope.sequenceNr shouldBe seqNr
         envelope.source shouldBe EnvelopeOrigin.SourceQuery
-        envelope.offset
+        envelope.offset.asInstanceOf[TimestampOffset]
       }
+      lastOffset1.timestamp shouldBe t1
+      lastOffset1.seen shouldBe Map(pid -> BufferSize)
 
       // process all events with backtracking query
-      (1 to BufferSize).foreach { seqNr =>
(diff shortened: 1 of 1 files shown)


core/src/test/scala/akka/persistence/r2dbc/query/EventsBySliceBacktrackingSpec.scala
```diff
@@ -387,6 +388,129 @@ class EventsBySliceBacktrackingSpec
...
+
+      val envelope = result.expectNext(10.seconds)
+      envelope.persistenceId shouldBe pid
+      envelope.sequenceNr shouldBe seqNr
+      envelope.source shouldBe EnvelopeOrigin.SourceQuery
+
+      result.cancel()
+    }
+
+    "not get stuck in backtracking when buffer-size events from several pids with same timestamp are all filtered" in {
```
patriknw: added one more test: https://github.com/akka/akka-persistence-r2dbc/pull/705/commits/1e1ae72b4eabdcba12a2b10e08df85c85291808f

the difference is that it's using more than one pid

not expecting that this would cover anything more, but it also fails without the fix

patriknw (approved):
LGTM

PR #709, opened 2025-11-29
fix: Bucket time slots must be rounded down

patriknw:
Illustrating the problem:
```
select 
  extract(EPOCH from db_timestamp) as a, 
  extract(EPOCH from db_timestamp)::BIGINT as b, 
  floor(extract(EPOCH from db_timestamp))::BIGINT as c 
  from journal
  where db_timestamp = '2025-04-30T15:23:39.906176Z' and seq_nr = 1785;
  
         a         |     b      |     c
-------------------+------------+------------
 1746026619.906176 | 1746026620 | 1746026619
 ```

This may result in that the counts fall into wrong buckets and then the slice query can stall on wrong end time.


ennru (approved):
I'm floored.
LGTM.

PR #717, opened 2026-01-19
fix: Snapshots as starting point for eventsBySlices, take 2

patriknw:
* The initial implementation suffered from two problems
  1. Not all entites have snapshots, and then it would miss those events when adjusting the start offset
  2. When emitting all snapshots first, followed by events, the offset storage can be two far ahead and events might be missed when restarting from stored offsets
* Instead, this implementation drives the stream from the ordinary eventsBySlices query.
  - reads snapshot sequence number on first occurance of a persistence id, and keeps that in memory
  - loads snapshot and emits when corresponding event sequence number is seen
  - skips event with sequence number lower than the snapshot sequence number
* Means that it must still read all events, but the benefit is that it doesn't have to process events before snapshots

See https://github.com/akka/akka-persistence-r2dbc/pull/410

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
@@ -9,12 +9,15 @@ import java.time.Instant
 import java.time.{ Duration => JDuration }
 import java.util.UUID
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicLong
+
 import scala.annotation.tailrec
 import scala.collection.immutable
 import scala.collection.mutable
 import scala.concurrent.Future
 import scala.concurrent.duration.Duration
 import scala.concurrent.duration.FiniteDuration
+
 import akka.NotUsed
 import akka.actor.ExtendedActorSystem
 import akka.actor.typed.pubsub.Topic
@@ -127,6 +130,8 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
   // key is tuple of entity type and slice
   private val heartbeatPersistenceIds = new ConcurrentHashMap[(String, Int), String]()
   private val heartbeatUuid = UUID.randomUUID().toString
+  // gaps are allowed for heartbeat sequence numbers, but increasing for each heartbeat pid (uuid makes it unique)
+  private val heartbeatSeqNr = new AtomicLong
   log.debug("Using heartbeat UUID [{}]", heartbeatUuid)
 
   // Optional caching of latestEventTimestamp results
@@ -206,7 +211,7 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
     new EventEnvelope(
       TimestampOffset(timestamp, Map.empty),
       heartbeatPersistenceId(entityType, slice),
-      1L,
+      heartbeatSeqNr.incrementAndGet(),
       eventOption = None,
       timestamp.toEpochMilli,
       _eventMetadata = None,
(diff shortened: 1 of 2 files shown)


ddl-scripts/create_tables_postgres.sql
```diff
@@ -42,9 +42,6 @@ CREATE TABLE IF NOT EXISTS snapshot(
   PRIMARY KEY(persistence_id)
 );
 
--- `snapshot_slice_idx` is only needed if the slice based queries are used together with snapshot as starting point
-CREATE INDEX IF NOT EXISTS snapshot_slice_idx ON snapshot(slice, entity_type, db_timestamp);
-
```
johanandren: Add something in the migration guide about dropping the index, or do we expect nobody was using this yet?

pvlugter:
> > Deeper change, but can we lazy load the event payloads, like for backtracking
> 
> That would be good for the catchup scenario, but probably not for the normal processing. Not sure we should complicate that with some adaptive hybrid approach. At least, let's try this first with some realistic data.

Start from snapshots is most useful for rebuilds, so optimising this further could make sense. If we did find that useful, could do something that combines something like the old approach with the new approach. Old approach did all the snapshots first, then regular event queries, which doesn't work. But we could take the idea of first doing an initial phase — where we process all the earlier events in a filtered lazy load way and the initial snapshots, until we pass the max snapshot timestamp — and then switch fully over to regular event processing after that. So rather than processing all snapshots and then start event processing from min snapshot timestamp, process catch up until max snapshot timestamp to initialise the projection.

core/src/main/scala/akka/persistence/r2dbc/internal/StartingFromSnapshotStage.scala
```diff
@@ -4,101 +4,178 @@
...
+        snapshotState.get(env.persistenceId) match {
+          case Some(s) =>
+            val eventIsAfterSnapshot = env.sequenceNr > s.seqNr
+
+            if (eventIsAfterSnapshot) {
+              // event is after snapshot, emit event
+              // we can't ignore it when snapshot is not emitted, because it might have been emitted in
+              // previous incarnation, but then the stream was restarted
+              push(out, env)
+            } else if (!s.emitted && env.sequenceNr == s.seqNr) {
```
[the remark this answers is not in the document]
patriknw: We as in Akka runtime/sdk don't delete events, until the whole entity is deleted, and then all events and snapshots are deleted, much later.

I'm not sure if snapshotting and retention strategies in EventSourcedBehavior can be setup to delete event seqNr == snapshot seqNr. However, deleting events combined with projections requires considerations anyway, and we have that documented somewhere.

core/src/main/scala/akka/persistence/r2dbc/internal/StartingFromSnapshotStage.scala
```diff
@@ -6,99 +6,212 @@ package akka.persistence.r2dbc.internal
...
+          inProgress = false
+          if (env.sequenceNr == seqNr) {
+            // snapshot should be emitted, load full snapshot
+            loadCorrespondingSnapshot(env)
+          } else if (env.sequenceNr > seqNr) {
+            // event is ahead of snapshot, emit event
+            updateState(env.persistenceId, seqNr, emitted = false)
+            push(env)
+          } else if (filterCount >= heartbeatAfter) {
+            pushHeartbeat()
```
patriknw: I have added heartbeats when many events have been filtered out. That will give observability progress and offset storage progress downstreams. I have to deal with these heartbeats in projections. As is, they would be like filtered events, and then probably handled as duplicates in offset store. https://github.com/akka/akka-persistence-r2dbc/pull/717/commits/8847b84533d5be71663b9a5a268a09dd0ce4ffc2

core/src/main/scala/akka/persistence/r2dbc/internal/StartingFromSnapshotStage.scala
```diff
@@ -6,99 +6,212 @@ package akka.persistence.r2dbc.internal
...
+          } else {
+            // snapshot will be emitted later, ignore event
+            updateState(env.persistenceId, seqNr, emitted = false)
+            ignore(env)
+          }
```
pvlugter: In places that we're pushing a heartbeat, rather than ignoring the envelope, should it also call `updateState`? Similar for `loadSnapshotCallback`. And also the update of the latest timestamp in ignore. Would seem that we should always do the same as ignoring/filtering, but also push the heartbeat. Could be simpler to not repeat logic, and move the heartbeat push into the `ignore` method?

pvlugter (approved):
LGTM. Updated approach looks good. Just a question around the state updates for heartbeats, and possible simplification of the logic here.

PR #725, opened 2026-04-24
feat: additional index and query for persistenceIds

aludwiko:
References https://github.com/lightbend/akka-runtime/issues/5060


aludwiko:
We decided to go with a different strategy: https://github.com/akka/akka-persistence-r2dbc/pull/726

PR #726, opened 2026-04-29
feat: currentPersistenceIdsBySlices to access recently active persistenceIds

aludwiko:
References https://github.com/lightbend/akka-runtime/issues/5060

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
@@ -260,7 +260,7 @@ private[r2dbc] class PostgresQueryDao(executorProvider: R2dbcExecutorProvider) e
         AND deleted = $sqlFalse
         ORDER BY persistence_id, db_timestamp DESC
       )
-      ORDER BY db_timestamp, persistence_id DESC
+      ORDER BY db_timestamp DESC, persistence_id ASC
       LIMIT ?;
       """
     }
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
@@ -247,15 +247,16 @@ private[r2dbc] class PostgresQueryDao(executorProvider: R2dbcExecutorProvider) e
   /**
    * the inner query has the ordering required by the distinct, the outer query applies desired order and a limit
    */
-  protected def currentPersistenceIdsBySlicesSql(minSlice: Int, maxSlice: Int): String =
-    sqlCache.get(minSlice, s"currentPersistenceIdsBySlicesSql-$minSlice-$maxSlice") {
+  protected def persistenceIdsBySlicesSql(toDbTimestampParam: Boolean, minSlice: Int, maxSlice: Int): String =
+    sqlCache.get(minSlice, s"persistenceIdsBySlicesSql-$minSlice-$maxSlice-$toDbTimestampParam") {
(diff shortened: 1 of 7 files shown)


core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
```diff
@@ -984,6 +984,44 @@ final class R2dbcReadJournal(system: ExtendedActorSystem, config: Config, cfgPat
       _.map(deserializeBySliceRow))
   }
 
+  /**
+   * Get the current active persistence ids for entities of `entityType` — those with at least one event whose
+   * `db_timestamp` is greater than or equal to the timestamp carried by `offset`, and whose slice is in the `[minSlice,
+   * maxSlice]` range. The supported `offset` is [[TimestampOffset]] and [[Offset.noOffset]] (which maps to
+   * `Instant.EPOCH`).
```
patriknw: shall we document the ordering? and that it can't be used for paging?

patriknw (commented):
looking good

core/src/main/scala/akka/persistence/r2dbc/query/scaladsl/R2dbcReadJournal.scala
johanandren: And that it doesn't work with sharded databases, or did I get that wrong?

patriknw: regarding sharded databases it should be documented in same way as eventsBySlices, i.e. it should work as long as the given slice range fits one database

aludwiko: Docs are now improved.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
```diff
@@ -244,6 +244,45 @@ private[r2dbc] class PostgresQueryDao(executorProvider: R2dbcExecutorProvider) e
...
+      val toDbTimestampCondition = if (toDbTimestampParam) "AND db_timestamp <= ?" else ""
+      sql"""
+      SELECT persistence_id FROM (
+        SELECT DISTINCT ON (persistence_id) persistence_id, db_timestamp
+        FROM ${journalTable(minSlice)}
+        WHERE entity_type = ?
+        AND ${sliceCondition(minSlice, maxSlice)}
+        AND db_timestamp >= ? $toDbTimestampCondition
+        AND deleted = $sqlFalse
+        ORDER BY persistence_id, db_timestamp DESC
```
patriknw, line 261: does this sub-select have the same number of rows as the outer? shall we add LIMIT here too, or is it clever enough to understand that?

patriknw (commented):
looking good

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
aludwiko: Yeah, I was in this place last week, so DISTINCT ON is not a grouping step first—it depends on the ordered stream of rows. In other words, it needs the full ordered dataset. That's why for large journals from-to should be more conservative. If we apply the limit in the inner query, we might miss some persistenceIds. 

Unless we truly want to set a hard limit, then the resulting list will be a best-effort approximation.

patriknw: I see, then we should at least always have a reasonable from/to interval when using this from akka-runtime

patriknw (approved):
LGTM

PR #730, opened 2026-05-06
chore: fixing persistenceIdsBySlicesSql for older Postgres versions

aludwiko:
An older version of Postgres (e.g. 15) will throw exception for a subquery without alias:
```
io.r2dbc.postgresql.ExceptionFactory$PostgresqlBadGrammarException: subquery in FROM must have an alias
```

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
@@ -259,7 +259,7 @@ private[r2dbc] class PostgresQueryDao(executorProvider: R2dbcExecutorProvider) e
         AND db_timestamp >= ? $toDbTimestampCondition
         AND deleted = $sqlFalse
         ORDER BY persistence_id, db_timestamp DESC
-      )
+      ) AS subquery
       ORDER BY db_timestamp DESC, persistence_id ASC
       LIMIT ?;
       """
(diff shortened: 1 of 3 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresQueryDao.scala
```diff
@@ -259,7 +259,7 @@ private[r2dbc] class PostgresQueryDao(executorProvider: R2dbcExecutorProvider) e
         AND db_timestamp >= ? $toDbTimestampCondition
         AND deleted = $sqlFalse
         ORDER BY persistence_id, db_timestamp DESC
-      )
+      ) AS subquery
```
patriknw, line 262: not sure what the error was, but are we missing test coverage?

patriknw: again, not reading when you link directly to the changes

patriknw: shall we add a nightly CI test with Postgres 15?

https://github.com/akka/akka-persistence-r2dbc/blob/main/.github/workflows/nightly.yml

aludwiko: Yeah the issue was only visible when testing with PG 15. We could set a fixed version in the docker compose file. PG 15 is from 2022.

aludwiko: > shall we add a nightly CI test with Postgres 15?

If you think it's worth it, in the end, the query type is quite specific.

johanandren: We don't really document anywhere what versions of Postgres we support as far as I can see. Maybe we should.

patriknw: add nightly CI for 15, since that is at least important for (some of) us

aludwiko: ok, added, only for some subset of options

johanandren (approved):

patriknw:
probably enough to test with one jdk version for postgres 15, let's use 25

PR #733, opened 2026-05-14
feat: persistenceIdsAndTimestampsBySlices query to include timestamps

pvlugter:
Having the timestamps for the latest event for these persistence IDs is very useful for creating queries when there are many IDs within a small time range. I'm currently experimenting with this for the akka console and CLI.


pvlugter:
This could still be useful at some point, but not going to use it currently for autonomous agents.

PR #737, opened 2026-06-24
chore: Update change date for 1.3.15

patriknw:


patriknw:
The SQL Server test fails repeatedly (3 times). First I suspected some regression from the additional column feature, but that is not the case. Other suspect is the Netty bumps and incompatibility with the sql server driver, but it passed in that bump PR. Probably just bad luck and same kind of flakiness of sql server driver as we have seen before.

PR #743, opened 2026-07-30
bump: R2DBC versions; explicit Netty override

ennru:
* bump: org.postgresql r2dbc-postgresql 1.1.2.RELEASE (was 1.1.1.RELEASE)
* bump: provided io.r2dbc r2dbc-mssql 1.0.5.RELEASE (was 1.0.4.RELEASE)
* explicit dependencies on Netty to bump its version within the 4.1 line past what Reactor Netty uses.

Added a safety net to ensure all Netty dependencies are of the version explicitly stated so that further changes keep those aligned.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -118,17 +118,21 @@ lazy val core = (project in file("core"))
 
       // match on the organization, other organizations publish artifacts named netty-* as well
       // netty-tcnative-* has its own version scheme.
-      val lagging = update.value.allModuleReports.filterNot(_.evicted).map(_.module).filter { module =>
-        module.organization == "io.netty" &&
+      val lagging = update.value.allModuleReports
+        .filterNot(_.evicted)
+        .map(_.module)
+        .filter { module =>
+          module.organization == "io.netty" &&
           !module.name.startsWith("netty-tcnative") &&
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -101,10 +103,35 @@ lazy val core = (project in file("core"))
   .settings(
     name := "akka-persistence-r2dbc",
     libraryDependencies ++= Dependencies.core,
+    // explicitly overriding Netty dependencies for downstream use
+    libraryDependencies ++= NettyModules.map("io.netty" % _ % NettyVersion),
+    dependencyOverrides ++= NettyModules.map("io.netty" % _ % NettyVersion),
```
ennru, line 108: I'm not confident this is the best solution. We need to set the Netty version even for downstream use, thus `libraryDepenencies` but with that alone the `managedClasspath` safety net still saw old Netty versions.

octonato: Do you mean, `libraryDependencies` is not enough therefore you needed `dependencyOverrides`?

octonato (approved):

build.sbt
ennru: Yes, exactly. That was unexpected.

PR #751, opened 2026-08-04
chore: Test native image on demand only; adapt release issue template

ennru:


johanandren:
How about we trigger it on dependency change instead, should be pretty easy now, just check if the bom has changed as a precondition.

johanandren:
Like this https://github.com/akka/akka-persistence-r2dbc/pull/752

PR #759, opened 2026-09-01
perf: batch snapshot-seqNr lookups for start-from-snapshot

johanandren:
## Summary

Adds an opt-in alternative to `StartingFromSnapshotStage` for `eventsBySlicesStartingFromSnapshots`. The original stage does one blocking, unpipelined DB round-trip per cache miss, which stalls the whole partition once the number of distinct persistence ids exceeds `cache-capacity`. `BatchingStartingFromSnapshotStage` batches misses into one round-trip per group and keeps pulling upstream while a batch is in flight.

Default is `start-from-snapshot.batching.enabled = false`. Behavior is unchanged unless it is turned on.

## Mechanism

 - `SnapshotDao.sequenceNumbersOfSnapshots` batches sequence-number lookups. The default falls back to concurrent per-id calls; `PostgresSnapshotDao` overrides it with a real `= ANY(?)` query grouped by data partition.
 - `BatchingStartingFromSnapshotStage` buffers envelopes in a single ordered queue and only ever decides from its head, so ordering matches the original stage.
 - H2 and SQL Server opt back into the concurrent fallback instead of inheriting the Postgres batched query.

## Testing

 - `BatchingStartingFromSnapshotStageSpec`: 5 unit tests, deterministic via hand-controlled `Promise`s, no database needed. Covers both races above plus ordering and the buffered-envelope bound.
 - Existing `EventsBySliceStartingFromSnapshotSpec` / `EventsByPersistenceIdStartingFromSnapshotSpec` pass unchanged with the default (off), and pass with batching forced on against real Postgres.
 - `BatchingSnapshotLookupBenchmark` (test sources, a plain `main`, not run by `sbt test`): seeds Postgres with a cache-miss-heavy dataset and measures throughput with batching on and off. Local runs show 25-45x.

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/BatchingStartingFromSnapshotStage.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/BatchingStartingFromSnapshotStage.scala
@@ -64,10 +64,6 @@ import akka.util.RecencyList
  * Loading the actual snapshot payload for entities whose boundary event is reached is left serialized as before — that
  * only happens once per entity (not once per cache eviction) so it isn't the dominant cost, but it is a natural next
  * step to pipeline the same way if it turns out to matter.
- *
- * Known gaps:
- *   - memory bound of the envelope buffer under pathological miss patterns not validated
- *   - snapshot-load step not pipelined (see above)
  */
 @InternalApi private[r2dbc] object BatchingStartingFromSnapshotStage {
   private case class SnapshotState(seqNr: Long, emitted: Boolean)
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/BatchingStartingFromSnapshotStage.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/BatchingStartingFromSnapshotStage.scala
@@ -29,41 +29,20 @@ import akka.util.RecencyList
 
 /**
  * Alternative to `StartingFromSnapshotStage` for workloads where a projection partition handles many more distinct
- * persistence ids than `cache-capacity`. In that situation the original stage's cache-miss rate approaches 1, and it
- * degrades into one fully serialized, blocking DB round-trip per event, because it does not pull the next upstream
- * element until the current lookup's callback has completed. Total pipeline throughput then becomes bound by round-trip
- * latency instead of DB or write capacity.
+ * persistence ids than `cache-capacity`. There, the original stage's cache-miss rate approaches 1, and it becomes fully
+ * serialized and round-trip-latency bound: it never pulls the next event until the current lookup completes.
  *
- * This stage keeps the same LRU cache, but resolves a batch of cache misses with a single round-trip instead of one
- * round-trip per miss. All envelopes — cache hits included — pass through a single ordered `pendingQueue` and are only
- * ever decided (push / ignore / load snapshot) strictly from its head, one at a time:
- *   - every envelope is enqueued in arrival order, and if its persistence id isn't cached its id is added to a
- *     pending-lookup set
- *   - upstream keeps being pulled (bounded by `maxBufferedEnvelopes`) to keep the queue filling
- *   - once `lookupBatchSize` distinct new ids have accumulated (anywhere in the queue, not just at the head),
- *     `maxBufferedEnvelopes` is reached, a linger timer expires, or upstream finishes, a *single* batched lookup
- *     (`SnapshotDao.sequenceNumbersOfSnapshots`) resolves the whole pending set in one round-trip
- *   - the queue is then advanced from the head for as long as each head's persistence id is resolved; advancing stops
- *     the moment the head's id isn't resolved yet (nothing later in the queue can be emitted before it without
- *     reordering) or a snapshot payload load is in flight for the current head
+ * This stage resolves cache misses in batches (`SnapshotDao.sequenceNumbersOfSnapshots`) instead of one round-trip per
+ * event, and keeps pulling upstream while a batch is in flight. All envelopes pass through a single ordered queue and
+ * are only ever decided from its head, so a cache hit can never be emitted ahead of an earlier, still unresolved
+ * envelope.
  *
- * Every envelope, hit or miss, must pass through the same head-of-queue gate: deciding a cache-hit envelope out of
- * order, ahead of an earlier envelope still waiting on a lookup or a snapshot load, would reorder the stream.
+ * A persistence id's cache entry is pinned against LRU eviction for as long as an envelope that depends on it, has not
+ * yet been decided, including while its snapshot payload is being loaded. See `loadCorrespondingSnapshot` for why the
+ * payload-load case needs the same protection.
  *
- * A persistence id that is resolved in the cache but still referenced by an envelope sitting in `pendingQueue` (stuck
- * behind an earlier, not-yet-resolved head) is pinned and exempt from LRU eviction until it is dequeued. Without this,
- * the entry could be evicted between being resolved and being dequeued, leaving that envelope permanently unresolved
- * and stalling the stage. The same applies while a snapshot payload load is in flight: the envelope that triggered it
- * has already been dequeued (and unpinned) by the time the load starts, but its persistence id must still not be
- * evicted for the load's duration - a batch lookup for a different id can run concurrently with the load (nothing gates
- * `triggerBatch` on `awaitingSnapshotLoad`), and if that eviction happened, a redelivery of the same id would see a
- * cache miss and start a redundant lookup racing the load, non-deterministically clobbering the `emitted` flag the load
- * is about to set and opening the door to a duplicate snapshot-envelope emission on a later redelivery.
- * `loadCorrespondingSnapshot` pins the id explicitly for exactly this reason.
- *
- * Loading the actual snapshot payload for entities whose boundary event is reached is left serialized as before — that
- * only happens once per entity (not once per cache eviction) so it isn't the dominant cost, but it is a natural next
- * step to pipeline the same way if it turns out to matter.
+ * Loading the snapshot payload itself stays serialized. That happens once per entity, not once per cache eviction, so
+ * it is not the dominant cost.
  */
 @InternalApi private[r2dbc] object BatchingStartingFromSnapshotStage {
   private case class SnapshotState(seqNr: Long, emitted: Boolean)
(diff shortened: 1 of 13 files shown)


patriknw:
looking good
----
Claude review:

## Verdict

The idea is sound and the ordering redesign (everything through one head-of-queue gate) is the right call. But there is **one blocking bug that hangs the stream permanently**, plus a dialect problem and a throughput nit.

---

## 1. Deadlock: a cache entry can be evicted between enqueue and decide

`BatchingStartingFromSnapshotStage.scala:263-266` — an id is added to `pendingLookupIds` **only** in `onPush`, and only when it's not in `snapshotState` at arrival time. But `snapshotState` is an LRU that keeps evicting while the envelope waits in `pendingQueue`.

Trace (cacheCapacity 3, lookupBatchSize 2, for brevity):

1. `onPush e1(X)` — X uncached → `pendingLookupIds = {X}`. `advance()` stops: head unresolved.
2. `onPush e2(A)` — A **is** cached → *not* added to `pendingLookupIds`. Queue `[e1(X), e2(A)]`. Head still X, so e2 is not decided.
3. `onPush e3(Y)` — `pendingLookupIds = {X, Y}` → batch fires.
4. `batchCallback` → `updateState(X, …)` evicts LRU tail → **A is evicted**; `updateState(Y, …)` evicts the next one.
5. `advance()`: e1(X) decided; head is now e2(A) → `snapshotState.get(A) == None` → `continue = false`.

A is gone from the cache and is in no lookup set. `advance()` (`:244`) has no re-request path, `maybeTriggerBatch` (`:226`) only looks at `pendingLookupIds`, and nothing else ever adds to it. `pendingQueue` fills to `maxBufferedEnvelopes`, `pullIfRoom` stops pulling, `completeIfDone` can never fire (`pendingQueue.isEmpty` is false forever). The stream hangs silently — no failure, no restart, projection just stops.

This isn't exotic with the defaults either. `recency.update` is only called on *resolve*, never on a cache hit (same as the original stage — but there it was harmless). So a frequently-hit pid decays toward the LRU tail regardless of how often it's hit, and any hit envelope whose entry sits within ~`lookupBatchSize` of the tail is a candidate. With `cache-capacity = 10000` / `lookup-batch-size = 200` that's on the order of a couple of percent *per cache-hit envelope* — and the whole point of the feature is workloads that churn the cache. It will fire.

The `reference.conf` note at `:203-205` spotted the eviction hazard but mis-diagnosed the consequence: it says it "forc[es] a redundant lookup", and there is no code path that does a redundant lookup — it stalls.

Fix, either:
- in `advance()`'s `None` branch, re-add the head's pid (`pendingLookupIds += pid; maybeTriggerBatch()`); needs an `inFlightIds` set to avoid re-querying an id already in the current batch, otherwise you get a duplicate query per stall; **or**
- pin ids that are present in `pendingQueue` so `updateState` can't evict them (cleaner — no wasted queries, and it makes the invariant "enqueued ⇒ resolvable" structural rather than something the batch sizing has to protect).

Either way `advance()` should also be reachable from `onUpstreamFinish`, which currently can't complete in this state either.

## 2. SQL Server and H2 inherit the Postgres `= ANY(?)` override

`SnapshotDao.scala:56-62` says the default falls back to single-id lookups "so existing dialects (H2, SQL Server) keep working without changes". That's not what happens: `H2SnapshotDao` and `SqlServerSnapshotDao` both `extends PostgresSnapshotDao`, so they inherit the `sequenceNumbersOfSnapshots` override at `PostgresSnapshotDao.scala:276` and never see the trait default.

For SQL Server that's a runtime failure — `= ANY(?)` isn't T-SQL, there's no array parameter type (`SqlServerTagsCodec` already joins tags into a delimited string for exactly that reason), and `SqlServerSnapshotDao` overrides `selectSeqNrSql` for named params but not `selectSeqNrsSql`. Needs an explicit override in `SqlServerSnapshotDao` delegating to the trait default (or a `STRING_SPLIT`/TVP form). H2 is likelier to work but should be confirmed on a real H2 rather than assumed.

## 3. Upstream pull and downstream drain don't overlap

`:271-277` — `onPull` only calls `pullIfRoom()` when `readyQueue` is empty. Once the buffer is full, `pullIfRoom` is a no-op, and while downstream drains N buffered envelopes there are zero upstream pulls; refill only starts after `readyQueue` empties, at which point downstream is idle. The two phases alternate instead of pipelining, which partly gives back what the batching bought. Call `pullIfRoom()` unconditionally after the push.


## Smaller notes

- Stage doc comment `:31` and `:64` are stale — it says "not wired into `R2dbcReadJournal` yet" and "config wiring … not added to reference.conf"; commit b65c0784 did both.
- `:33` names a customer ("the observed Westpac issue") in a public OSS source file. Worth replacing with a description of the workload shape.
- No tests, as the comment admits. `StartingFromSnapshotStageSpec` is parameterizable enough to cover both stages; the two cases I'd add first are the eviction-mid-queue case from §1 and an ordering case with a snapshot load in flight (the bug you already hit against real Postgres).
- `require(maxBufferedEnvelopes >= lookupBatchSize)` at `:90` but nothing relates `lookupBatchSize` to `cacheCapacity`. A batch larger than the cache self-evicts within a single `batchCallback` loop. Worth a `require` once §1 is fixed, or moot if you go with pinning.
- Minor: while a batch is in flight, a new envelope for an id already in that batch re-enters `pendingLookupIds` (`:263`), producing a redundant query and delaying `completeIfDone` by one linger. An `inFlightIds` set fixes this and §1 together.
- `sequenceNumbersOfSnapshots` taking an implicit `ExecutionContext` is out of step with the rest of the trait — every implementation already has one in scope.

The table-name grouping in `PostgresSnapshotDao` is correct, and the comment explaining why it isn't grouped by `executorFor` is worth keeping — table name is a function of `dataPartition(slice)`, which is finer than or equal to the connection-factory grouping, so batches never span a connection.

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresSnapshotDao.scala
```diff
@@ -259,6 +268,36 @@ private[r2dbc] class PostgresSnapshotDao(executorProvider: R2dbcExecutorProvider
         _.get[java.lang.Long]("seq_nr", classOf[java.lang.Long]))
   }
 
+  // One `= ANY(?)` query per data-partition instead of one query per id.
+  override def sequenceNumbersOfSnapshots(persistenceIds: Set[String])(implicit
+      ec: ExecutionContext): Future[Map[String, Long]] = {
```
patriknw: why does this have ExecutionContext parameter instead of using the existing 
```
implicit protected val ec: ExecutionContext = executorProvider.ec
```

core/src/main/scala/akka/persistence/r2dbc/internal/h2/H2SnapshotDao.scala
```diff
@@ -25,6 +28,12 @@ private[r2dbc] final class H2SnapshotDao(executorProvider: R2dbcExecutorProvider
 
   private val sqlCache = Sql.Cache(settings.numberOfDataPartitions > 1)
 
+  // The `= ANY(?)` batched query in PostgresSnapshotDao is not verified to work against H2's r2dbc driver, so opt
+  // back into the safe per-id fallback rather than inheriting the Postgres override.
+  override def sequenceNumbersOfSnapshots(persistenceIds: Set[String])(implicit
+      ec: ExecutionContext): Future[Map[String, Long]] =
+    sequenceNumbersOfSnapshotsConcurrently(persistenceIds)
```
patriknw, line 33: Will this fire 200 concurrent lookups? Those will be queued by the much smaller connection pool size, and might hit the `acquire-timeout`. Shall we simply not support batching for h2?

core/src/test/scala/akka/persistence/r2dbc/query/BatchingSnapshotLookupBenchmark.scala
```diff
@@ -0,0 +1,187 @@
...
+      akka.persistence.r2dbc.journal.publish-events = off
+      """)
+      .withFallback(TestConfig.config)
+
+  private val batchedConfig: Config =
+    ConfigFactory
+      .parseString("akka.persistence.r2dbc.query.start-from-snapshot.batching.enabled = true")
+      .withFallback(baselineConfig)
+
+  def main(args: Array[String]): Unit = {
```
patriknw, line 67: We don't have anything that covers the batching SQL in CI. Can we have some variant of the existing StartFromSnapshotSpec that has batching enabled (or same test with two configs)?

core/src/main/scala/akka/persistence/r2dbc/internal/BatchingStartingFromSnapshotStage.scala
```diff
@@ -0,0 +1,310 @@
...
+          pendingLookupIds.clear()
+          idsInFlight ++= ids
+          batchInFlight = true
+          sequenceNumbersOfSnapshots(ids)
+            .map(results => (ids, results))(ExecutionContext.parasitic)
+            .onComplete(batchCallback.invoke)
+        }
+      }
+
+      private def maybeTriggerBatch(): Unit = {
```
patriknw, line 237: Add `upstreamFinished` to the immediate-trigger condition? Once upstream is done, no further ids can arrive.

```scala
private def maybeTriggerBatch(): Unit = {
  if (!batchInFlight && pendingLookupIds.nonEmpty) {
    // once upstream is finished no further ids can arrive, so there is nothing to linger for
    if (upstreamFinished || pendingLookupIds.size >= lookupBatchSize || pendingQueue.size >= maxBufferedEnvelopes)
      triggerBatch()
    else if (!isTimerActive(BatchLingerTimerKey))
      scheduleOnce(BatchLingerTimerKey, batchLinger)
  }
}
```

and remove redundant guard in
```scala
override def onUpstreamFinish(): Unit = {
  upstreamFinished = true
  maybeTriggerBatch()
  advance()
  completeIfDone()
}
```

patriknw (commented):
looking good

core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresSnapshotDao.scala
johanandren: Sorted

core/src/main/scala/akka/persistence/r2dbc/internal/BatchingStartingFromSnapshotStage.scala
johanandren: Sorted [dc4a0b7](https://github.com/akka/akka-persistence-r2dbc/pull/759/commits/dc4a0b71cc333fb6bec33c798b1862681dbedf61)

core/src/main/scala/akka/persistence/r2dbc/internal/h2/H2SnapshotDao.scala
johanandren: Chunked into groups that are run one at a time now, chunk size half of connection pool size

core/src/test/scala/akka/persistence/r2dbc/query/BatchingSnapshotLookupBenchmark.scala
johanandren: Test added

patriknw (approved):
LGTM

johanandren:
Let's see what our test replay run shows (tomorrow) before merging this.

johanandren:
Did not turn out to be the root cause we thought it was, but valuable enough to include anyway.

PR #765, opened 2026-09-10
fix: Retrieve more bucket counts directly when catching up

patriknw:
* eventsBySlices queries could run without the `db_timestamp <= ?` upper bound when catching up a long backlog, resulting in expensive sequential scans
* one bucket count query covers at most ~28 hours, which a catching up projection can consume well within the 60 seconds before the next bucket count query was allowed
* a time range with too few events didn't give any upper bound either
* now the next buckets are retrieved immediately when the previous bucket count query was limited by its time range or number of buckets, rather than by current time
* and it continues directly with the next time range if there were too few events
* the 60 seconds interval still applies once caught up
* BySliceQueryEstimateTimeRangeSpec reproduces it with an in-memory dao

### DIFF
--- a/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
+++ b/core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
@@ -119,6 +119,20 @@ import org.slf4j.Logger
     val Limit = 10000
 
     final case class Bucket(startTime: EpochSeconds, count: Count)
+
+    /**
+     * Upper bound (inclusive) of the time range for the bucket count query. The time range corresponds to `limit`
+     * number of buckets, or until `now` if that is earlier. No time range limit if `fromTimestamp` is `Instant.EPOCH`.
+     */
+    def countBucketsToTimestamp(fromTimestamp: Instant, limit: Int, now: Instant): Instant = {
+      if (fromTimestamp == Instant.EPOCH)
+        now
+      else {
+        // max buckets, just to have some upper bound
+        val t = fromTimestamp.plusSeconds(BucketDurationSeconds * limit + BucketDurationSeconds)
+        if (t.isAfter(now)) now else t
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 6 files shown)


core/src/main/scala/akka/persistence/r2dbc/internal/postgres/PostgresDurableStateDao.scala
```diff
@@ -872,15 +872,7 @@ private[r2dbc] class PostgresDurableStateDao(executorProvider: R2dbcExecutorProv
...
-    val toTimestamp = {
-      if (fromTimestamp == Instant.EPOCH)
-        now
-      else {
-        // max buckets, just to have some upper bound
-        val t = fromTimestamp.plusSeconds(Buckets.BucketDurationSeconds * limit + Buckets.BucketDurationSeconds)
-        if (t.isAfter(now)) now else t
-      }
-    }
+    val toTimestamp = Buckets.countBucketsToTimestamp(fromTimestamp, limit, now)
```
patriknw, line 876: I think we have more to fix for this in durable state, but let's take that as follow up

core/src/main/scala/akka/persistence/r2dbc/internal/BySliceQuery.scala
```diff
@@ -664,27 +683,50 @@ import org.slf4j.Logger
...
-            log.debug(
-              "{} retrieved [{}] event count buckets, with a total of [{}], from time [{}]",
-              logPrefix,
-              counts.size,
-              sum,
-              fromTimestamp)
-          }
-          newState
+      // the dao uses the same time range, but possibly with a slightly later `now`
+      val now = InstantFactory.now()
```
johanandren, line 687: The DAO methods (`PostgresQueryDao.countBuckets`, `PostgresDurableStateDao.countBuckets`) independently call `InstantFactory.now()` inside the query to compute their own `toTimestamp`. Under enough latency/clock skew between the two `now()` calls, the caller's reconstruction could occasionally disagree with what the DAO actually did?

patriknw:
sqlserver failing as usual, rerunning

johanandren (approved):

johanandren:
Tested with actual data, showed excellent results. Let's merge and release.

patriknw:
I will follow up with some more on this, but fine to release this first

PR #768, opened 2026-09-14
fix: Durable state retrieve more bucket counts directly

patriknw:
Follow up to https://github.com/akka/akka-persistence-r2dbc/pull/765

* Retrieve more bucket counts directly when catching up, in same way as for eventsBySlices
* Stale starting point: the durable-state bucket query started from `latestBacktracking`, which stops moving once backtracking is switched off because the stream is far behind. This is the durable-state version of what #641 fixed for events
* fix older rounding bug for Durable State
* Cleanup


johanandren (approved):
LGTM!

PR #769, opened 2026-09-15
bump: h2 2.5.250

ennru:


johanandren (approved):
LGTM, seems fine for the parts we have test coverage for. 

I'm still a little bit worried about diverging from the h2 version (esp bumping minor) that R2DBC-H2 is built against.

johanandren:
For the record, last upstream release R2DBC-H2 was 11 months ago, no bump for H2 PR:ed or merged since.

